PR #12, opened 2016-10-09
Issue 8/move lower phase hash

micylt:
`lower_phase_hash` is now included as a top level parameter in `sign_verification_record`. Also being appended to hashed_items separately from `verification_info`.

`phase_1` lower phase hash is being set to 0 since there's not a phase lower than `phase_1`.
`phase_3`  is keeping the list of `phase_2` `lower_phase_hashes` in addition to creating a separate `lower_phase_hash` from the hash of `lower_phase_hashes`,  which is included as a top level parameter.
--Took note to consider storing a structure of (`signatory`, `hash`) rather than `lower_phase_hashes` in `phase_3`.

Ran and tested the code after making the changes through `phase_4` without issues.

### DIFF
--- a/blockchain/processing.py
+++ b/blockchain/processing.py
@@ -29,13 +29,13 @@ __version__ = "2.0"
 __maintainer__ = "Joe Roets"
 __email__ = "joe@dragonchain.org"
 
-from blockchain.block import Block,               \
-                             BLOCK_FIXATE_OFFSET, \
-                             BLOCK_INTERVAL,      \
-                             get_block_time,      \
-                             get_current_block_id
+from blockchain.block import Block, \
+    BLOCK_FIXATE_OFFSET, \
+    BLOCK_INTERVAL, \
+    get_block_time, \
+    get_current_block_id
 
-from blockchain.util.crypto import valid_transaction_sig, sign_verification_record, validate_verification_record
+from blockchain.util.crypto import valid_transaction_sig, sign_verification_record, validate_verification_record, deep_hash
 
 from blockchain.util.thrift_conversions import thrift_record_to_dict, thrift_transaction_to_dict
 
@@ -99,6 +99,7 @@ class ProcessingNode(object):
             observer["callback"](config=observer["config"], **kwargs)
 
     """ INTERNAL METHODS """
+
     def _init_networking(self):
         """ currently only starts up network, may add more features """
         print 'initializing network'
@@ -250,9 +251,12 @@ class ProcessingNode(object):
             prior_block_hash = self.get_prior_hash(origin_id, phase)
             verification_info = approved_transactions
 
+            lower_phase_hash = str(deep_hash(0))
+
             # sign approved transactions
             block_info = sign_verification_record(signatory,
                                                   prior_block_hash,
+                                                  lower_phase_hash,
                                                   self.service_config['public_key'],
                                                   self.service_config['private_key'],
                                                   current_block_id,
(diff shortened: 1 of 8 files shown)


thrift/messaging.thrift
```diff
@@ -110,46 +110,47 @@ struct Node {
...
 struct Phase_1_msg {
-    1: VerificationRecordInfo record,
+    1: VerificationRecordCommonInfo record,
     2: list<Transaction> transactions
 }
 
 struct Phase_2_msg {
-    1: VerificationRecordInfo record,
+    1: VerificationRecordCommonInfo record,
     2: string lower_phase_hash,
```
j0j0r0: This isn't needed (it's in common struct now)

thrift/messaging.thrift
```diff
@@ -110,46 +110,47 @@ struct Node {
...
     2: list<string> lower_phase_hashes,
     3: i32 p2_count,
     4: list<string> business_list,
     5: list<string> deploy_loc_list
 }
 
 struct Phase_4_msg {
-    1: VerificationRecordInfo record,
+    1: VerificationRecordCommonInfo record,
     2: string lower_phase_hash
```
j0j0r0: Not needed (in common)

thrift/messaging.thrift
```diff
@@ -110,46 +110,47 @@ struct Node {
...
     5: string business,
     6: string deploy_location
 }
 
-/* rename business_list to businesses and deploy_loc_list to deploy_locations */
+/* TODO: rename business_list and deploy_loc_list to businesses and deploy_locations respectively */
 struct Phase_3_msg {
-    1: VerificationRecordInfo record,
+    1: VerificationRecordCommonInfo record,
     2: list<string> lower_phase_hashes,
```
j0j0r0: Cool - this is where the later tuple or array will be (signatory_id, hash)

thrift/messaging.thrift
```diff
@@ -110,46 +110,47 @@ struct Node {
...
 struct Phase_4_msg {
-    1: VerificationRecordInfo record,
+    1: VerificationRecordCommonInfo record,
     2: string lower_phase_hash
 }
 
 union Phase_5_msg {
     1: Transaction transaction,
-    2: VerificationRecordInfo record,
+    2: VerificationRecordCommonInfo record,
```
j0j0r0: Hadn't noticed this before (phase 5 is still in dev) but I think that this should maybe go away since there is no generic full verification record message that we could put in here.
As it is, this would only contain the common info, and not enough to do the job.
Leave this for now as it's UNRELATED to this ticket - I will create a new issue and explain further.

blockchain/processing.py
```diff
@@ -307,17 +311,20 @@ def _execute_phase_2(self, config, phase_1_info):
 
             valid_transactions, invalid_transactions = self.check_tx_requirements(phase_1_info.transactions)
 
+            # TODO: take lower_phase_hash out of verification_info and have it hashed in sign_verification instead
```
j0j0r0: This is done yes? Am I missing something, or should this TODO go away?

thrift/messaging.thrift
j0j0r0: Ticket #16 created.

blockchain/processing.py
micylt: Yes, it's done. I will blow it away today. Weird, I thought I took the lower_phase_hash out of those phase messages. I will make sure they're out today.

thrift/messaging.thrift
micylt: Fixed.

thrift/messaging.thrift
micylt: Fixed.

thrift/messaging.thrift
micylt: Fixed.

blockchain/processing.py
micylt: Fixed.

PR #15, opened 2016-10-10
Issue 3/statuscode function created

wolfpack94:
Created a function set_txn_status function in transaction_svc.py, taking in the log and the error code and properly logging the corresponding error code statuses in the log file.


j0j0r0:
I will wait for closeout of prior PR which have their commits included in this PR.
:(

PR #33, opened 2016-10-12
partial work - bitcoin timestamping

lontivero:
This is the **unfinished-and-untested** bitcoin timestamping feature. 

What's done and what is not:
- `blockchain/timestamping.py` contains the _bitcoin timestamper_ that uses a local bitcoin node for managing the wallet (`getting utxos`, `getting new addresses`, `signing transactions`, etc), broadcasting transactions, querying transactions and tx outputs. It also contains an estimated fee provider that uses a 3rd party service for getting real-time estimated fee required for miners to include our tx in next block. All **what is in this file is tested**. This was tested with the `regtest` bitcoin network and also with the broadcasted transaction mined (in a block) as well as the transaction in the mempool.
- `blockchain/processing.py` was modified in order to implement phase 5 that receives a phase 4 message, verifies it and persists it as "pending for timestamping". There is a `cron` configured for running _every 30 seconds_ that performs a query against the db to get all the _"pending for timestamping" records_, builds the merkletree with those and timestamps the merkle tree root hash in the bitcoin blockchain. After that it updates all the records with the chainpoint proof receipt. So, nodes can verify the proof of existence for any record using the `BitcoinTimestamper` class.  This is not tested and it is not validated for correctness.
- The rest is probably not okay (or simply wrong), incomplete and untested.

### DIFF
--- a/blockchain/db/postgres/timestamp_db.py
+++ b/blockchain/db/postgres/timestamp_db.py
@@ -1,0 +1,96 @@
+"""
+Copyright 2016 Disney Connected and Advanced Technologies
+
+Licensed under the Apache License, Version 2.0 (the "Apache License")
+with the following modification; you may not use this file except in
+compliance with the Apache License and the following modification to it:
+Section 6. Trademarks. is deleted and replaced with:
+
+     6. Trademarks. This License does not grant permission to use the trade
+        names, trademarks, service marks, or product names of the Licensor
+        and its affiliates, except as required to comply with Section 4(c) of
+        the License and to reproduce the content of the NOTICE file.
+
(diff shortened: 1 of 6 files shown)


lontivero:
I would like to add that there are some 3rd party services that expose a RPC interface that can be used instead of having to install and maintain a local bitcoin node. blochain.info provides the RPC services [here is the documentatio](https://blockchain.info/es/api/json_rpc_api)

Voxelot:
@j0j0r0 should this be closed since #105 is under review?

j0j0r0:
Yes it should. Thank you.

PR #40, opened 2016-10-17
issue-34/isolate thrift dependencies

micylt:
All thrift conversions are now being done in network.py. Ran and tested up to phase 4 without issues.

May have to relook at issue-37 (refactoring lower_phase_hash) due to code being moved around.

### DIFF
--- a/blockchain/network.py
+++ b/blockchain/network.py
@@ -539,7 +539,16 @@ class BlockchainServiceHandler:
         self.connection_manager.processing_node.notify(3, phase_2_info=phase_2_info)
 
     def phase_3_message(self, phase_3):
-        self.connection_manager.processing_node.notify(4, phase_3_info=phase_3)
+        phase_3_info = {
+            'record': thrift_record_to_dict(phase_3.record),
+            'verification_info': {
+                'lower_phase_hashes': phase_3.lower_phase_hashes,
+                'p2_count': phase_3.p2_count,
+                'business_list': phase_3.business_list,
+                'deploy_location_list': phase_3.deploy_loc_list
+            }
+        }
+        self.connection_manager.processing_node.notify(4, phase_3_info=phase_3_info)
 
     def phase_4_message(self, phase_4):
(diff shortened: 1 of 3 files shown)


blockchain/processing.py
```diff
@@ -296,20 +296,20 @@ def _execute_phase_2(self, config, phase_1_info):
           - tx minus the payload
         """
         phase = 2
-        prior_block_hash = self.get_prior_hash(phase_1_info.record.origin_id, phase)
-        p1_verification_info = map(thrift_transaction_to_dict, phase_1_info.transactions)
+        phase_1_record = phase_1_info["record"]
+        p1_verification_info = phase_1_info["verification_info"]
+        phase_1_record['verification_info'] = p1_verification_info
+        prior_block_hash = self.get_prior_hash(phase_1_record["origin_id"], phase)
```
Voxelot: Maybe pull these field names into constants?

blockchain/network.py
```diff
@@ -518,13 +520,35 @@ def submit_verification(self, block_id, phase, block_record):
...
+                'business': phase_2.business,
+                'deploy_location': phase_2.deploy_location
+            }
+        }
+        self.connection_manager.processing_node.notify(3, phase_2_info=phase_2_info)
 
     def phase_3_message(self, phase_3):
-        self.connection_manager.processing_node.notify(4, phase_3_info=phase_3)
+        phase_3_info = {
+            'record': thrift_record_to_dict(phase_3.record),
```
Voxelot: See my other comment, we should start using constants for these fields.

Voxelot:
I'll open a separate issue for using constants in our messaging.

blockchain/network.py
micylt: Thank you for the feedback. I will create constants for these fields and look for anywhere else we should be using constants.

Voxelot:
Also in the future, you should use the GitHub issue link feature in your original comment when you make a pull request.

PR #41, opened 2016-10-17
proof of X impl

lontivero:
I migrated this file from the old private repo just for historical reasons and because i hope it would be useful **for discussing** about different ways to implement the _proof-of-something_ feature. I added more comments to the previous version.
# The idea

The idea that I show here is based on two components:
- **Proof**: an object that contains all the required data for proving that something was done. In case of PoW it containg the data for proving a hash less than a target was found after increasing a nonce several times. In case of PoB it contains a reference to the spent transaction output (burned coin). A proof has to be part of each block. 
- **Providers**: a components that 'provides' the proof. In the case of PoW a better name for this componente could be 'miner', in the case of PoB a better name could be 'burner' or 'spender'; but proof_provider is abstract name that could be good for all.
# Notes

It worth to note that these abractractions could not fit well for other kinds of proofs. It is also important to have in mind that some design desitions will be affected by the implementation of any of the PoX alternatives, for example: 
- if we decide to go for PoW then we have to review the 5 seconds block interval because mining time is not predictable.
- if we go for PoB then we need to have a native currency and a way for distribute. This could be also valid for PoS.

This means that PoX is not something easily 'plugable' because it affects other componentes and depends on (or affects to) other design desitions. Anyway, PoW is by sure the one that requires less changes and the only one that does not require a native currency in order to work.

### DIFF
--- a/blockchain/pox.py
+++ b/blockchain/pox.py
@@ -1,0 +1,186 @@
+"""
+Copyright 2016 Disney Connected and Advanced Technologies
+
+Licensed under the Apache License, Version 2.0 (the "Apache License")
+with the following modification; you may not use this file except in
+compliance with the Apache License and the following modification to it:
+Section 6. Trademarks. is deleted and replaced with:
+
+     6. Trademarks. This License does not grant permission to use the trade
+        names, trademarks, service marks, or product names of the Licensor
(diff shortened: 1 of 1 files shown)


cheeseandcereal:
stale

Issue #42, opened 2016-10-17
Use constants for message data structures


Voxelot:
There are currently a lot of dictionary structures being passed between the processing and network modules and the fields are being accessed via in place string literals. These should be replaced with constants to make refactoring easier and reduce potential bugs.

cyR3X:
I'll take a crack at this one if you would like...

micylt:
@cyR3X Absolutely. I added several global variables for some of the commonly used string literals, but there's still strings that should be global. Thank you.

cyR3X:
@micylt What are your thoughts on me adding a "constants.py" file to hold all of the constants/variables for the project?  It would need to be imported (either completely or partially) into each existing file, but long-term might be a good idea...  Please let me know your thoughts.

micylt:
@cyR3X That sounds like a good idea to me. We would obviously have to make sure the constants module is being included where it needs to be. Write something up if you'd like and we'll review it :). Let me know if you have anymore questions.

cyR3X:
@micylt I have made some changes, but cannot push them to the "Master" branch from my local repo...  Can you (or someone else) add me as a contributor?  Thank you.

PR #45, opened 2016-10-22
Issue 42/Add-constants

micylt:
issues #42. Added constants in various files where appropriate. Ran and tested without issues.


Voxelot:
It's a start but there's still a lot of magic strings in use.

PR #53, opened 2016-10-26
Remove uneeded legacy version support

JeremyRubin:
There seems to be no public versions with non-pluralized table names, so may as well remove this legacy support now.


j0j0r0:
Sorry for the spam - was on mobile, and couldn't get the merge...

PR #55, opened 2016-10-26
Hard Fork: switch to sign_digest to use intended hash

JeremyRubin:
Otherwise I believe current implementation falls back to sha1 for message digests of your hash, which is pretty bad!

### DIFF
--- a/blockchain/util/crypto.py
+++ b/blockchain/util/crypto.py
@@ -104,7 +104,7 @@ def sign_transaction(signatory,
 
     # sign merged hash
     log.info("Signing merged hash")
-    signature = ecdsa_signing_key.sign(str(merged_hash))
+    signature = ecdsa_signing_key.sign_digest(str(merged_hash))
     log.info("Base64 encoding the signature")
     digest = signature.encode('base64')
 
@@ -170,7 +170,7 @@ def sign_verification_record(signatory,
 
     verification_hash = final_hash(hashed_items)
 
-    signature = ecdsa_signing_key.sign(verification_hash)
+    signature = ecdsa_signing_key.sign_digest(verification_hash)
     digest = signature.encode('base64')
 
     verification_record = {
@@ -262,9 +262,9 @@ def validate_signature(signature_block, log=logging.getLogger(__name__)):
     # checking stripped hash if this is a transaction signature
     if "stripped_hash" in signature_block and signature_block['stripped_hash']:
         merged_hash = merge_hashes(signature_block["stripped_hash"], signature_block["hash"])
-        verifying_key.verify(decoded_digest, str(merged_hash))
+        verifying_key.verify_digest(decoded_digest, str(merged_hash))
     else:
-        verifying_key.verify(decoded_digest, str(signature_block["hash"]))
+        verifying_key.verify_digest(decoded_digest, str(signature_block["hash"]))
 
 
 def validate_verification_record(record, verification_info, log=logging.getLogger(__name__)):


JeremyRubin:
I should note, this is untested

micylt:
I will test this today.

j0j0r0:
Don't have time to look, but what's the difference between sign and sign_digest? 

@micylt @JeremyRubin  we're already making our own deterministic hash, so what is mean by the sha1 comment?

j0j0r0:
Sorry, mobile only right now. 
I may have time to look this evening.

JeremyRubin:
Ok so this has to do with another design thing you should change as well.

First off, you should restrict your curves to be ONLY SECP256k, right now you can read from an pem and it will use whatever curve is specified there. This restriction is a soft fork.

Sign takes an arbitrary string and sign_digest takes something you are asserting to be a cryptographic hash. If you don't use sign_digest, sign will hash whatever you pass it for you, using sha1.

The digest must be the size of the curve, you're using sha512, change it to sha256 and use secp256k1. The reason your code currently works is your sha512 is actually reduced to sha1 (160 bit) which is a small enough digest to work with NIST224p.

Voxelot:
Hey Jeremy, thanks for the catch on `sign` vs `sign_digest`. I wouldn't say we "fallback" to SHA1, as we're currently doing a SHA512 to generate our own digest before signing. However, from an audit standpoint, you're right, it'd be harder to verify the signature without using our crypto implementation.

What is the justification for using secp256k over prime256v? The latter is included in NSA suite B and standard across all Linux distributions, whereas secp256k may require a custom build of OpenSSL to generate.

In regards to a hard-fork, I don't see how allowing nodes to use higher levels of encryption would impact our blockchain. I agree that there should be some minimum requirements, but as long as clients verifying signatures are able to properly read and use the provided public key I don't forsee any issues.

JeremyRubin:
I think there's basically no point in using sha512 if you end up doing sha1 on the final signature. For the sake of simplicity, let's imagine i have a byte array for the sha512 hash of a tx, h = [b63, b62...b0].  What you're doing is effectively only signing h[:20]. (sha1 is only 160 bits). But actually, your story is worse than that! You're using sha1 which is (probably?) a weaker hash than truncated sha512.


W.r.t. restricting signatures, secp256k1 should already be usable on your platform today with the way signatures are implemented, so I'm not adding new behavior just restricting it. secp256k1 has a lot of review and a much faster implementation so if scalability is a concern than it's probably your best bet. Furthermore, secp256k1's parameters were selected by a predictable process, which is less likely to have backdoors than the way the nsa curves were generated. Using secp256k1 in sign_digest mode also requires you to use sha256, or at least a 256-bit length hash (you can use a longer one but I think it will be truncated? not sure exact behavior)

JeremyRubin:
I think rather than frame this as "as long as clients verifying signatures are able to properly read and use the provided public key" 

ask

Why would a dragonchain user want to use more than one type of signature? Why would a (future) dragonchain developer want to maintain support for prime192v1, secp224r1,
prime256v1, secp384r1, secp521r1, and secp256k1? Why would we want to later add restrictions on signature type when the python ecdsa library adds more signature types?

JeremyRubin:
You guys should really fix this, especially as SHA-1 is "broken"! But I'm probably not going to have time to work on this further, so closing.

j0j0r0:
Hi Jeremy - apologies.
Intent is to fix this, but there were architectural considerations surrounding whether we would require all nodes to have the exact same signing behavior (L1 nodes having much independence) - and is more than I have time to describe here.
I think that those questions are mostly resolved in the interest of simplicity (good), so I'd like to reopen - (I'll take ownership).

For my part, I had trouble finding documentation on `sign_digest` - would you have any reference for that?

Thank you.

JeremyRubin:
Just read the source.

https://github.com/warner/python-ecdsa/blob/master/src/ecdsa/keys.py#L234

I would even go one more and recommend sign_digest_deterministic, but that's up to you (and non-consensus).

j0j0r0:
Perfect - will look into it.
Thank you @JeremyRubin!

Quasso:
@j0j0r0 any update on this status of this? I would love to hear.

Quasso:
@j0j0r0 any update on the status of this? I would love to hear.

cryptudutcch:
Soooooooo, we good?

JeremyRubin:
No this is still not addressed.

Still should be fixed, but [now that you have a network launched?] it requires a hard fork to fully fix.

j0j0r0:
Apologies for the late reply - we actually had something that conflicted internally way, way back on this that kept us from doing what we should - I'll get someone assigned to take a look.

cheeseandcereal:
stale

Issue #28, opened 2016-10-11
query criteria


j0j0r0:
Add owner (and other fields) field to the query service criteria.
Add ability to query JSON payload.

j0j0r0:
Add query capability for all transaction fields not already implemented.

For timestamp (`create_ts` and `transaction_ts`) fields, use the following encoding:

1. For range, translate integer dash integer as inclusive range, e.g. `120987-129089` should translate to `timestamp >= 120987 AND timestamp <= 120989`
1. For older than, use dash integer e.g. -120987 should translate to `timestamp <= 120987`
1. For newer than, use `integer` `dash` e.g. `-120987` should translate to `timestamp >= 120987`
1. For no range query, use `integer` e.g. `120987` should translate to `timestamp = 120987`

All ranges should result in inclusive queries.

j0j0r0:
We should split the JSON query task into another ticket.

wolfpack94:
Has already been merged. Closing issue.

PR #56, opened 2016-10-29
Issue 28/query criteria

wolfpack94:
Issue #28 Added query services for all header fields. 
Can be queried by running query_svc.py and navigating to http://localhost:8080/transaction/?header_field=value also implemented is querying by timestamp. Note: All timestamps must be unix epoch timecode to work when querying. localhost:8080/transaction/?create_ts=1475155787.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -97,6 +97,9 @@ logs/
 # pki
 pki/
 
+# .DS_Store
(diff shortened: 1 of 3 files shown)


Voxelot:
Looks good :+1:

PR #57, opened 2016-10-29
Issue 20/phase5 definition change

micylt:
issue #20 

`Phase_5_msg` in `messaging.thrift` is now a union that accepts either a `Transaction`, `VerificationRecord`, `hash` or a `miscellaneous string`.

`VerificationRecord` is a union that can either be a phase 1, 2, 3 or 4 `thrift` message type.

Implemented a public broadcast function in `network` that takes in a block, sets `VerificationRecord` and `phase_5_msg` and sends the block out to all known phase 5 nodes for `public transmission`.

Added public transmission on/off switches for all phases to the `blockchain.yml` network config. i.e if a config is set to `on` for phase 2, that block will be sent for `public transmission` by all phase 2 nodes it is sent to. If it's set to `off`, it is not sent for `public transmission`. Each phase checks for public transmission switches.

Added `public transmission` structure to the `VerificationRecordCommonInfo`.

Moved some logic that's now being used in several places into functions.

Refactored a few things.

Ran and tested several times without issues.

### DIFF
--- a/blockchain/network.py
+++ b/blockchain/network.py
@@ -133,7 +133,6 @@ class ConnectionManager(object):
         self.port = port
         self.business = None
         self.deploy_location = None
-        self.public_transmission = None
         # phase_type => {nodes} (dictionary of connected nodes)
         self.peer_dict = {}
         self.config = None
@@ -406,6 +405,17 @@ class ConnectionManager(object):
 
     def phase_1_broadcast(self, block_info, phase_type):
         """ sends phase_1 information for phase_2 execution """
+        phase_1_msg = self.get_p1_message(block_info)
+
+        for node in self.peer_dict[phase_type]:
+            try:
+                node.client.phase_1_message(phase_1_msg)
+            except:
+                logger().warning('failed to submit to node %s', node.node_id)
+                continue
+
+    # returns thrift phase 1 message structure
+    def get_p1_message(self, block_info):
         record = block_info['verification_record']
         transactions = map(convert_to_thrift_transaction, record['verification_info'])
         verification_record = get_verification_record(record)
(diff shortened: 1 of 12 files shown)


micylt:
Resolved conflicts.

* Note: currently nodes sends verification record to all know phase 5s nodes it knows about for public transmission in the current implementation. I believe we want the record to only be sent off to one phase 5 node for public transmission.

j0j0r0:
Regarding how many P5 nodes to which it should broadcast, it should probably be configurable.

A node will need to be able to control where it's sent based upon budget factors as well since the P5 nodes will likely be charging for the service.

For now, yes, let's either leave it or add in a simple stop at one.

micylt:
Resolved conflicts with master. Ran and tested without issues, ready to merge.

j0j0r0:
Thank you @micylt

Voxelot:
@micylt @j0j0r0: please take a look at code review comments.

blockchain/network.py
```diff
@@ -75,21 +74,19 @@
...
-VERIFICATION_INFO = 'verification_info'
-
 PHASE_1_NODE = 0b00001
 PHASE_2_NODE = 0b00010
 PHASE_3_NODE = 0b00100
 PHASE_4_NODE = 0b01000
 PHASE_5_NODE = 0b10000
 
-DATABASE_NAME = os.getenv(pg.ENV_DATABASE_NAME, pg.DEFAULT_DB_NAME)
+DATABASE_NAME = os.environ.get('BLOCKCHAIN_DB_NAME')
```
micylt: That should not have been taken out. It should be 
DATABASE_NAME = os.getenv(pg.ENV_DATABASE_NAME, pg.DEFAULT_DB_NAME)

blockchain/processing.py
```diff
@@ -358,7 +357,7 @@ def _execute_phase_2(self, config, phase_1_info):
                 self.network.public_broadcast(block_info, phase)
 
             # send block info for phase 3 validation
-            self.network.send_block(self.network.phase_2_broadcast, block_info, phase_1_record['phase'])
+                self.network.send_block(self.network.phase_2_broadcast, block_info, phase_1_record[PHASE])
```
micylt: No.

PR #62, opened 2016-11-04
Improved clarity on public/private key generation in Readme

mikedeshazer:
Providing clarity in the public/private key generation process to create less friction on setup for developers stepping through the Readme file


j0j0r0:
Merged.
Sorry for the delay.

PR #66, opened 2016-11-07
Docker

mbachtel:
This is not working with my current docker install but want to get the merge in place so that others can further verify or fix.


j0j0r0:
@Voxelot is reviewing.

Voxelot:
Refactored this into a new docker-compose branch #72

PR #69, opened 2016-11-08
adding Content-Type header of application/json in get txn service res…

Adam-Bronfin:
GET /transaction service was defaulting Content-Type of response to HTML

Added line to explicitly set Content-Type to "application/json"

**Also adding content type on response of POST /transaction service


Voxelot:
Looks good 👍

PR #75, opened 2016-11-12
init commit -- adding payload hash directly onto txn table for added …

Adam-Bronfin:
Adding payload hash directly onto transaction table.

Having this extra field will enable some future capabilities, such as allowing a user prove a given payload/document exists on the blockchain. This can be done by hashing it themselves and confirming that a corresponding transaction exists.

### DIFF
--- a/blockchain/db/postgres/transaction_db.py
+++ b/blockchain/db/postgres/transaction_db.py
@@ -54,13 +54,14 @@ SQL_INSERT = """INSERT into transactions (
                             family_of_business,
                             line_of_business,
                             payload,
+                            payload_hash,
                             signature,
                             owner,
                             transaction_type,
                             status,
                             actor,
                             entity
-                          ) VALUES  (%s, to_timestamp(%s), to_timestamp(%s), %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)"""
+                          ) VALUES  (%s, to_timestamp(%s), to_timestamp(%s), %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)"""
 SQL_UPDATE = """UPDATE transactions SET status = %s, block_id = %s WHERE transaction_id = %s"""
 SQL_FIXATE_BLOCK = """UPDATE transactions SET status='pending', block_id=%i WHERE status = 'new' AND transaction_ts >= to_timestamp(%i) AND transaction_ts <= to_timestamp(%i)"""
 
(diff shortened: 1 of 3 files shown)


blockchain/db/postgres/transaction_db.py
```diff
@@ -228,6 +229,7 @@ def insert_transaction(txn):
         header["family_of_business"],
         header["line_of_business"],
         Json(txn["payload"]),
+        txn["payload_hash"],
```
j0j0r0: I think that we don't want the payload hash at txn top level.

blockchain/db/postgres/transaction_db.py
```diff
@@ -54,13 +54,14 @@
                             family_of_business,
                             line_of_business,
                             payload,
+                            payload_hash,
```
j0j0r0: Want to discuss payload hash being in the table separate from the other hashes (full and stripped hashes are in the signature structure, stored as JSON type).

While this makes sense for the use case of querying for the hash of an arbitrary payload, it may suit us to store payload_hash in sig and duplicate it here along with the other hashese.

blockchain/util/crypto.py
```diff
@@ -90,10 +90,14 @@ def sign_transaction(signatory,
...
-        hashed_items.append(deep_hash(transaction["payload"]))
-        # generate hash with with hashed payload included
+        payload_hash = deep_hash(transaction["payload"])
+        hashed_items.append(payload_hash)
+        # generate hash with hashed payload included
         log.info("Generating hash")
         hash = final_hash(hashed_items)
+
+        # add payload hash to transaction so it can be stored in transaction table
+        transaction["payload_hash"] = payload_hash
```
j0j0r0: We don't want the payload hash to be top level in the txn.

Should probably only exist alongside the other hashes in the sig block.

sql/txpool.sql
```diff
@@ -75,6 +75,8 @@ CREATE TABLE IF NOT EXISTS transactions (
     /* what are a few examples? */
     payload JSON,
 
+    payload_hash VARCHAR(500),
```
j0j0r0: Should this be 500?

j0j0r0:
@Adam-Bronfin considering to drop this change - want to discuss in person especially as it relates to the notary function?

Basic concerns: 

1. **Security issue** with having payload (business data) hash available standalone (e.g. on a simple payload structure it would be easy to brute force exposure of data)
1. Consistency - it seems that we would need the payload hash in the sig block alongside the other hashes, but also flattened with heading data for efficient query, but then are the other hashes needed in the table?

j0j0r0:
Going to close this PR.
We can talk Monday or early next week to see where you are with the other notary functionality and if we need to reopen this.

Issue #65, opened 2016-11-05
Broadcast Receipts (data synchronization)


micylt:
Implement node receipt functionality. Currently verification records are sent up the pipe to phase 2, 3, etc. but there is nothing in place for say a phase 2 to send back a signed verification recored receipt to phase 1. In other words, once any phase validates and signs a verification record, we want that node to send the signed verification record back to the phase that sent it.

Design:
Phase/node X sends record to phase Y; the next time phase X sends a record to phase Y, phase Y should send back all signed records it has on record for phase X, including records Y has for the X from nodes Z, A, B, etc. Ideally we would like all info being sent between nodes to be done on the same pipe. There should be a receipt cut-off point. If node Y has not sent back a receipt for a verification record by this point, we may want to have a configuration that says drop node Y (bad node) at that point.

micylt:
@j0j0r0 and @Voxelot Please review and let me know if anything should be added/removed/modified. Thank you.

j0j0r0:
Only thing that I would add is that the phase Y node should send back all signed records that it has for the calling node **that have not previously been returned to the calling node**

1. Node X sends record to Node Y
1. Node Y sends back all signed records it has for Node X, including records that it has signed, as well as any higher phases that it has received from other nodes

I expect that we will need to add a few fields to the datastore - I think in a new table (`broadcast_receipt`):

- broadcast attempts - count of attempted broadcasts to higher phased node(s) - for a broadcast to multiple phases (e.g. Phase 2), this should only be counted as 1 
- receipt confirmed (Boolean field to indicate that the signed record has been returned down to the calling node)

_I think that we need a new table because the broadcast attempts count to multiple nodes will need to be tracked individually for retry and to determine if a node is not answering. Potentially, we could store the attempts in a structured field, but I think that we'll want something more efficient for the query_

Any node that is not signing a Node X's record after some configured number of attempts can be marked as `bad` or `suspected_bad` and the network module can act to replace it.

If the next level has not been attained, then the processing module can continue to use nodes that are arranged by the network module to move forward.

Let me know if you have any questions.

Voxelot:
I think eagerly returning upstream verification records could be a useful optimization, but it doesn't guarantee completeness. For the initial implementation, let's start with a simple cron-based pull engine. 

[...]

PR #77, opened 2016-12-19
Issue 65/data synchronization

micylt:
Issue #65

* Now on successful phase broadcasts, the requested node returns unsent verification ids to the calling node. The calling node then checks for which ids it has or has not yet received from the returned list of requested ids. The calling node then makes a request to get back records it has not yet received. At this point the calling node is also sending back what data it has received, this is what handles duplicate data. 

* The requested node sends back verification records (receipts) not yet sent and marks them as sent. After the calling node receives the receipts, they are stored and that nodes transfer table is updated for replication. This is what handles receipts from later phases. the requested nodes "last_transfer_time" is also updated.

* A cron timer has been added to check the last transfer time for all connected nodes. If the last transfer time is greater than x time, a request is made for receipts from said node.

* A new table has been added for verification transfers with the following fields:
    `origin_id`
    `transfer_to` => node to send back to
    `verification_id`
    `sent` bit boolean

* Database queries have been added for transfer table.

* A few upgrades and improvements have been added to the block_verification table.

* Minor changes have been made to the thrift api for returning verification ids.

* Moved various thrift conversions from network into thrift_conversions api.


blockchain/network.py
```diff
@@ -670,6 +688,41 @@ def create_node_from_peer(peer):
...
+        node_unreceived = []
+        all_guids = received + unreceived
+
+        # retrieve ids that calling node is claiming as unreceived
+        for guid in unreceived:
+            node_unreceived += self.get_unsent_transfer_ids(transfer_to=None, verification_id=guid)
+
+        # mark all verifications received as sent
+        for guid in all_guids:
+            vr_transfers_db.set_verification_sent(guid)
```
Voxelot: it should only mark the guid as sent for the requesting (`transfer_to`) node!!! The it is currently will prevent the verification guid from going to more than one node after the verification has already been transferred!  Only one p2 node would receive the p3 verification record because all the transfer records would be marked as sent for the other p2s. Also, what if an attacker made transfer_data calls to your p2+ nodes with arbitrary verification ids as already sent? The lower nodes would never receive them.

blockchain/network.py
```diff
@@ -670,6 +688,41 @@ def create_node_from_peer(peer):
...
+        return unsent_transfer_ids
+
+    def transfer_data(self, received, unreceived):
+        """ mark verifications as sent and return unsent verifications """
+        verifications = []
+        node_unreceived = []
+        all_guids = received + unreceived
+
+        # retrieve ids that calling node is claiming as unreceived
+        for guid in unreceived:
```
Voxelot: since the guids are already the primary keys, there's no need to fetch them from the transfer table first just to return the same guid back. You can take this call out.

blockchain/network.py
```diff
@@ -670,6 +688,41 @@ def create_node_from_peer(peer):
...
+            node_unreceived += self.get_unsent_transfer_ids(transfer_to=None, verification_id=guid)
+
+        # mark all verifications received as sent
+        for guid in all_guids:
+            vr_transfers_db.set_verification_sent(guid)
+        # retrieve unreceived records
+        for guid in node_unreceived:
```
Voxelot: should just be `for guid in unreceived:` to avoid all the extra db calls I mentioned earlier.

blockchain/db/postgres/vr_transfers_db.py
```diff
@@ -0,0 +1,119 @@
...
+GET_VERIFIED_RECORDS = """SELECT * FROM vr_transfers"""
+SQL_MARK_RECORD = """UPDATE vr_transfers SET sent = B'1' WHERE verification_id = %s"""
+SQL_INSERT_QUERY = """INSERT INTO vr_transfers (
+                                  origin_id,
+                                  transfer_to,
+                                  verification_id
+                                ) VALUES (%s, %s, %s)"""
+
+
+def get_unsent_verification_records(node_transmit_id=None, verification_id=None):
```
Voxelot: See comments below in the networking section. This method only needs the node_transmit_id param.

blockchain/network.py
```diff
@@ -670,6 +688,41 @@ def create_node_from_peer(peer):
 
         return map(create_node_from_peer, self.connection_manager.connections)
 
+    def get_unsent_transfer_ids(self, transfer_to=None, verification_id=None):
```
Voxelot: only needs the transfer_to param

thrift/messaging.thrift
```diff
@@ -192,7 +192,7 @@ service BlockchainService {
 
    list<string> receipt_request(1: string signatory),
 
-   list<VerificationRecord> transfer_data(1: list<string> received, 2: list<string> unreceived),
+   list<VerificationRecord> transfer_data(1: string transfer_to, 2: list<string> received, 3: list<string> unreceived),
```
Voxelot: Could we figure out the `transfer_to` id based on the pass phrase used when the connection is registered instead of passing in any arbitrary node id?

blockchain/network.py
```diff
@@ -700,22 +700,22 @@ def get_unsent_transfer_ids(self, transfer_to=None, verification_id=None):
...
 
-    def transfer_data(self, received, unreceived):
+    def transfer_data(self, transfer_to, received, unreceived):
         """ mark verifications as sent and return unsent verifications """
         verifications = []
         node_unreceived = []
         all_guids = received + unreceived
 
         # retrieve ids that calling node is claiming as unreceived
         for guid in unreceived:
```
Voxelot: again, why are we querying the transfer table for ids we already have? I think we should get rid of the node_unreceived array altogether and just use the` unreceived` list when retrieving unreceived records below.

blockchain/network.py
```diff
@@ -700,22 +700,22 @@ def get_unsent_transfer_ids(self, transfer_to=None, verification_id=None):
...
-            node_unreceived += self.get_unsent_transfer_ids(transfer_to=None, verification_id=guid)
+            node_unreceived += self.get_unsent_transfer_ids(transfer_to=transfer_to, verification_id=guid)
 
         # mark all verifications received as sent
         for guid in all_guids:
-            vr_transfers_db.set_verification_sent(guid)
+            vr_transfers_db.set_verification_sent(transfer_to, guid)
         # retrieve unreceived records
         for guid in node_unreceived:
```
Voxelot: change `node_unreceived` -> `unreceived`

blockchain/network.py
```diff
@@ -700,22 +700,22 @@ def get_unsent_transfer_ids(self, transfer_to=None, verification_id=None):
...
+    def transfer_data(self, transfer_to, received, unreceived):
         """ mark verifications as sent and return unsent verifications """
         verifications = []
         node_unreceived = []
         all_guids = received + unreceived
 
         # retrieve ids that calling node is claiming as unreceived
         for guid in unreceived:
-            node_unreceived += self.get_unsent_transfer_ids(transfer_to=None, verification_id=guid)
+            node_unreceived += self.get_unsent_transfer_ids(transfer_to=transfer_to, verification_id=guid)
```
Voxelot: can we change the `get_unsent_transfer_ids` api to only use the `transfer_to` param? I don't see why we need the `verification_id` param.

blockchain/db/postgres/verification_db.py
```diff
@@ -120,7 +137,7 @@ def get_records(block_id, origin_id, phase):
         get_connection_pool().putconn(conn)
 
 
-def get_all(block_id, limit=None, offset=None, phase=None):
+def get_all(block_id, limit=None, offset=None, phase=None, origin_id=None):
     # Build query
     query = SQL_GET_ALL
     if block_id:
```
micylt: Fixed.

blockchain/db/postgres/verification_db.py
```diff
@@ -160,9 +184,38 @@ def get_all(block_id, limit=None, offset=None, phase=None):
         get_connection_pool().putconn(conn)
 
 
-def insert_verification(verification_record):
+def get_all_replication(block_id, phase, origin_id):
```
micylt: Fixed. Let me know if you think anything should be added/changed.

sql/b_trans.sql
```diff
@@ -0,0 +1,67 @@
+/*
```
micylt: Fixed. Originally we thought this would stand for 'block transfer'. Renamed to ver_transfers.

blockchain/network.py
micylt: Fixed.

blockchain/network.py
```diff
@@ -670,6 +688,41 @@ def create_node_from_peer(peer):
...
+        # retrieve ids that calling node is claiming as unreceived
+        for guid in unreceived:
+            node_unreceived += self.get_unsent_transfer_ids(transfer_to=None, verification_id=guid)
+
+        # mark all verifications received as sent
+        for guid in all_guids:
+            vr_transfers_db.set_verification_sent(guid)
+        # retrieve unreceived records
+        for guid in node_unreceived:
+            verifications += verification_db.get_verifications(guid)
```
micylt: Fixed.

blockchain/network.py
```diff
@@ -409,81 +410,84 @@ def send_block(self, phase_broadcast, block_info, phase):
...
-        return phase_3_msg
-
     # TODO: implement this broadcast
     def phase_4_broadcast(self, block_info, phase_type):
         pass
 
+    def timed_receipt_request(self):
+        """ time based receipt request """
+        for node in self.connections:
+            if int(time.time()) - node.last_transfer_time >= 900:
```
micylt: Fixed.

blockchain/network.py
micylt: Fixed.

blockchain/network.py
micylt: Fixed.

blockchain/network.py
micylt: Fixed. I am assuming we still want to set 'all_guids' as sent.

blockchain/db/postgres/vr_transfers_db.py
micylt: Fixed.

blockchain/network.py
micylt: Fixed.

blockchain/network.py
micylt: Fixed.

blockchain/network.py
micylt: Fixed.

blockchain/db/postgres/verification_db.py
```diff
@@ -59,6 +59,23 @@ def get_cursor_name():
     return str(uuid.uuid4())
 
 
+def get_verifications(verification_id):
```
micylt: Fixed.

blockchain/processing.py
```diff
@@ -456,7 +463,11 @@ def _execute_phase_3(self, config, phase_2_info):
                                                       )
 
                 # inserting verification info after signing
-                verfication_db.insert_verification(block_info['verification_record'])
+                verification_id = str(uuid.uuid4())
+                verification_db.insert_verification(block_info['verification_record'], verification_id)
+
+                # inserting receipt of signed verification for data transfer
+                vr_transfers_db.insert_transfer(phase_2_record['origin_id'], phase_2_record['signature']['signatory'], verification_id)
```
micylt: Fixed.

blockchain/processing.py
```diff
@@ -429,6 +429,7 @@ def _execute_phase_3(self, config, phase_2_info):
             # storing valid verification record
             verification_db.insert_verification(phase_2_record)
 
+            # retrieve all phase 2 records for current block
             phase_2_records = self.get_sig_records(phase_2_record)
```
Voxelot: I'd move this line below 438, that way we only fetch the p2 records when the reqs are met and they're needed for the transfer records.

blockchain/block.py
```diff
@@ -71,7 +71,7 @@ def get_block_id(secs_since_epoch):
 
 
 def get_current_block_id():
-    # get the seconds in 5 sec intervals
+    # get_verifications the seconds in 5 sec intervals
```
Voxelot: accidental find/replace?

blockchain/db/postgres/transaction_db.py
```diff
@@ -263,7 +263,7 @@ def update_transaction(txn):
 
 
 def fixate_block(start_ts_range, end_ts_range, block_id):
-    # get all tx within the previous block
+    # get_verifications all tx within the previous block
```
Voxelot: find/replace bug?

blockchain/block.py
micylt: Yes, find/replace. Fixed.

blockchain/db/postgres/transaction_db.py
micylt: Yes, find/replace. Fixed.

blockchain/processing.py
micylt: @Voxelot and @j0j0r0 I think we need to keep this where it is. This is where we retrieve the records to check that the diversity requirements have been met.

thrift/messaging.thrift
micylt: Done. Also updated `receipt_request` to use `pass_phrase` as well.

blockchain/util/crypto.py
```diff
@@ -35,7 +35,7 @@
 
 import time
 import logging
-logging.basicConfig()
+# logging.basicConfig()
```
j0j0r0: Do we need to retain this?

sql/ver_transfers.sql
```diff
@@ -0,0 +1,67 @@
+/*
```
j0j0r0: Can we rename the file to `vr_transfers` per the table name?

blockchain/db/postgres/verification_db.py
```diff
@@ -120,20 +120,28 @@ def get_records(block_id, origin_id, phase):
...
 
 
-def get_all(block_id, limit=None, offset=None, phase=None):
-    # Build query
+def get_all(limit=None, offset=None, **params):
+    """ return all verification records matching given parameters """
     query = SQL_GET_ALL
-    if block_id:
-        query += """ WHERE block_id = """ + str(block_id)
+    multi_param = False
```
micylt: Done.

blockchain/db/postgres/verification_db.py
```diff
@@ -120,20 +120,28 @@ def get_records(block_id, origin_id, phase):
...
-        query += """ phase = """ + str(phase)
+        query += """ phase = """ + str(params["phase"])
+        multi_param = True
 
-    query += """ ORDER BY block_id DESC """
+    if "origin_id" in params:
+        if multi_param:
+            query += """ AND """
+        query += """ origin_id = """ + str(params["origin_id"])
+        multi_param = True
```
micylt: @j0j0r0 Added this in the event that we add more parameters later on. Want me to take it out?

blockchain/util/crypto.py
micylt: Removed.

sql/ver_transfers.sql
micylt: Done.

sql/b_trans.sql
```diff
@@ -0,0 +1,67 @@
...
+/* Don't drop tables unless you really want to */
+CREATE TABLE IF NOT EXISTS vr_transfers (
+    /* Original owner of the transaction data (Blockchain ID) */
+    origin_id VARCHAR(256),
+    /* The node to transmit this record to */
+    transfer_to VARCHAR(256),
+    /* verification record id */
+    verification_id UUID,
+    /* Indicates whether a record has been transmitted or not */
+    sent BIT(1) DEFAULT b'0'
```
j0j0r0: One more question, why is this a BIT type and not boolean? 

Are we expecting to encode something more later?

micylt: Did it this was as per the design in the ticket. Perhaps this was done for efficiency reasons? Good question for @Voxelot.

blockchain/network.py
```diff
@@ -692,26 +688,16 @@ def create_node_from_peer(peer):
...
-        except:
-            logger().warning("An SQL error has occurred.")
-
-        return unsent_transfer_ids
-
-    def transfer_data(self, transfer_to, received, unreceived):
+    def transfer_data(self, pass_phrase, received, unreceived):
         """ mark verifications as sent and return unsent verifications """
+        self.authorize_pass_phrase(pass_phrase)
+        transfer_node = self.registered_nodes[pass_phrase]
```
Voxelot: Add a `None` check just in case the pass_phrase is bogus.

blockchain/db/postgres/verification_db.py
Voxelot: Will do.

sql/b_trans.sql
Voxelot: Ah my bad, boolean would be a better fit.

sql/vr_transfers.sql
```diff
@@ -0,0 +1,67 @@
...
+    /* Original owner of the transaction data (Blockchain ID) */
+    origin_id VARCHAR(256),
+    /* The node to transmit this record to */
+    transfer_to VARCHAR(256),
+    /* verification record id */
+    verification_id UUID,
+    /* Indicates whether a record has been transmitted or not */
+    sent BIT(1) DEFAULT b'0'
+);
+COMMIT;
```
Voxelot: One thing I left out of the design which will be needed is an index for the query that fetches unsent records.

```
BEGIN;
CREATE INDEX vr_transfer_tts_idx ON test2 (transfer_to, sent);
COMMIT;
```

j0j0r0: +1

sql/b_trans.sql
micylt: Done.

blockchain/network.py
micylt: Done.

sql/vr_transfers.sql
micylt: Done.

micylt:
Ready to merge.

j0j0r0:
Good stuff!

Issue #78, opened 2016-12-23
update-query_svc-readme


wolfpack94:
Update the query_svc execution instructions in the main dragonchain readme to correct parameters and running instructions.

Voxelot:
PR merged. Closing.

PR #79, opened 2016-12-23
Updated README.md

wolfpack94:
Issue #78 There was a support question regarding the query_svc.py execution instructions in the README, so I updated that section with all the necessary information to run query services from the command line. I also added my name and linkedin to the bottom of the contributors section, can be removed if requested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ If you have Apache Thrift installed you can regenerate these classes by using th
     
 ## Query Service
 
-    python <Dragonchain Home>/blockchain/query_svc.py -p port //defaults to 8080
+    python <Dragonchain Home>/blockchain/query_svc.py [-p port] //defaults to 8080
 
 ### Example Query
 
--- a/README.md
+++ b/README.md
@@ -74,14 +74,8 @@ If you have Apache Thrift installed you can regenerate these classes by using th
 # Execution
 
 ## Set the PYTHONPATH variable
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -73,14 +73,29 @@ If you have Apache Thrift installed you can regenerate these classes by using th
...
+        PYTHONPATH="/path/to/dragonchain"
+
 ## Transaction Service
 
     python <Dragonchain Home>/blockchain/transaction_svc.py --private-key <Dragonchain Home>/pki/sk.pem --public-key <Dragonchain Home>/pki/pk.pem
     
 ## Query Service
 
-    python <Dragonchain Home>/blockchain/query_svc.py --private-key <Dragonchain Home>/pki/sk.pem --public-key <Dragonchain Home>/pki/pk.pem
+    python <Dragonchain Home>/blockchain/query_svc.py -p port //defaults to 8080
```
Voxelot: Let's try making these look more like standard CLI instructions. See Argparse for example (https://docs.python.org/3/library/argparse.html).

I'm thinking something like `query_svc.py [-p port]` would be more appropriate.

j0j0r0:
Also, on the PYTHOHPATH variable, both Linux and Mac should use export (make it a single line for Linux/Mac.

wolfpack94:
Alright, didn't realize it was the same for linux as well. Made the modifications.

Issue #84, opened 2016-12-30
Automate thrift generation via shell script


wolfpack94:
Write a shell script in order to generate the thrift files, copy the select ones into the blockchain/gen folder and delete the gen-py folder from the thrift directory.

Voxelot:
Is this really necessary? Don't we already have a way to do this in [setup.py](https://github.com/dragonchain/dragonchain/blob/master/setup.py#L60)?

wolfpack94:
Yes you are correct, running that will do the same but when running the command: thrift -r --gen py ../thrift/messaging.thrift you do not need all the files it generates and so this script only copies over what is necessary (i.e. ttypes.py, BlockchainService.py, BlockchainService-remote. when running the command above, you have to manually copy over the recreated files. It might not be needed, just thought it would be quicker.

PR #85, opened 2016-12-30
added shell script to automate thrift generation

wolfpack94:
Issue #84 Wrote and verified script worked for generating thrift files, copying Blockchain-Service.py, BlockchainService-remote and ttypes.py into the blockchain/gen folder and finally deleting the gen-py folder from the thrift generation.


Voxelot:
See my comment [here](https://github.com/dragonchain/dragonchain/issues/84#issuecomment-269833659)

Issue #25, opened 2016-10-11
Bitcoin Bridge Implementation


j0j0r0:
Implement a bridge to allow interaction between Bitcoin and internal currencies.

Need ability to track addresses (internal and external will be Bitcoin addresses) and fire events based upon activity.

Need ability to create Bitcoin transactions and broadcast to Bitcoin network.

Will probably involve integration of open source bitcoin wallet software.

wolfpack94:
`thrift`
struct public_proof {
   1: VerificationRecord verification_record
}

wolfpack94:
`struct public_proof {
     1: VerificationRecord verification_record
  }`

wolfpack94:
I modified the verification_info created in _execute_timestamping so that the json struction now looks like this, 
`verification_info = {`
`         verification_records = {`
`                   origin_id = {`
`                              verification_id = hash`
`                   }`
`          }`
`}`

is that what you were talking about on Thursday @j0j0r0 ?

j0j0r0:
Yes, exactly. 
:)

j0j0r0:
Although you may have to refresh my memory about the outer verification_info...

wolfpack94:
Updated the code snippet to the full `verification_info` structure as it stands now in the code.

wolfpack94:
@j0j0r0 @Voxelot Can this be closed since it was merged?

Issue #87, opened 2017-01-06
Optimize Phase 5 Broadcast Selection 


micylt:
Currently records are being sent to all known phase 5 nodes for `public broadcast`.

Records should only be sent to one phase 5 node. Ideally, records should be sent to the healthiest phase 5 node.

Health being based on node latency, uptime, etc.

PR #105, opened 2017-03-18
Issue 25/bitcoin bridge

wolfpack94:
Issue #25 
Bitcoin Bridge Implementation
This is a continuation of the work from Lucas that adds the ability to send the data being sent up the Dragonchain onto the public Bitcoin network.

- Can embed hashes into a bitcoin payload which is pushed onto the bitcoin network with a small mining fee.

- Can either be connected to a local or remote bitcoin node depending on your needs. 

Modules added:
- timestamping.py
- timestamp_db.py


wolfpack94:
Fixing conflicts right now.

micylt:
Resolved conflicts. Ran, tested and ready to merge.

j0j0r0:
Ok. I'll review asap and talk to you tomorrow about it @micylt 

Thanks.

docker/custom-docker-compose.yml
```diff
@@ -0,0 +1,66 @@
+version: '2'
```
j0j0r0: Is this file only for your environment?
Did you mean to add it?

sql/timestamp.sql
```diff
@@ -0,0 +1,61 @@
...
+    timestamp_id UUID PRIMARY KEY,
+
+    block_id INT,
+
+    origin_id VARCHAR(256),
+
+    /* confirm: when the transaction record was created */
+    create_ts timestamptz,
+
+    /* chainpoint receipt containing the required data for verifying the timestamp */
```
j0j0r0: I think that I've answered this on my own - used in query for inclusion in new record for timestamp operation.

thrift/messaging.thrift
```diff
@@ -154,13 +155,12 @@ union VerificationRecord {
...
-    2: VerificationRecord verification_record,
-    /* Level 5 node will NOT hash this field */
-    3: string hash,
-    /* Level 5 node WILL hash this field */
-    4: string misc
+union Phase_5_request {
+    1: VerificationRecord verification_record
+}
+
+struct Phase_5_msg { /* change to Public_5_msg */
```
j0j0r0: Is `Phase_5_msg` still used?
I see a reference in the `gen` code, but I thought it was replaced by `Phase_5_request`?

README.md
```diff
@@ -125,6 +125,33 @@ Pre-create a directory for log files before execution.
     --host        (defaults to localhost)
     --port, -p    (defaults to 8080)
     --phase       (required)
+    
+    For phase 5 node only:
+    --bitcoin-network (mainnet/testnet/regtest)
+    
+## Running a phase 5 node
```
wolfpack94: Thank you.

sql/timestamp.sql
wolfpack94: Changed the comment to make more sense

blockchain/processing.py
```diff
@@ -39,17 +39,23 @@
 
 from blockchain.util.crypto import valid_transaction_sig, sign_verification_record, validate_verification_record, sign_subscription, final_hash
 
-from db.postgres import postgres
-from db.postgres import transaction_db
-from db.postgres import verification_db
-from db.postgres import vr_transfers_db
+from bitcoin.core import *
```
wolfpack94: As mentioned today over the phone, the requirements.txt file was modified to include this library and version. Just clarifying for pull request sake.

blockchain/processing.py
```diff
@@ -166,6 +173,17 @@ def trigger_handler():
 
         elif config[PHASE] == 5:
             self._add_registration(5, self._execute_phase_5, config)
+            self._add_registration("timestamp", self._execute_timestamping, config)
```
wolfpack94: The first registration is the actual phase 5 function which retrieves records from previous phases. The second registration is the chron for timestamping which happens every 30 seconds? i believe. I might not have that correct but I believe that is what it is.

blockchain/processing.py
```diff
@@ -166,6 +173,17 @@ def trigger_handler():
...
+
+            # Setup phase 5 cron
+            trigger = CronTrigger(second='*/30')
```
wolfpack94: Might have to explain the config override. I can change but I don't fully understand what is being asked.

blockchain/processing.py
```diff
@@ -606,9 +626,85 @@ def _execute_phase_4(self, config, phase_3_info):
...
-        print "phase 5 executed"
+        phase = 5
+        verification_record = verification['record']
+
+        verification_info = verification['verification_info']
+        verification_record['verification_info'] = verification_info
+
+        if validate_verification_record(verification_record, verification_info):
+            timestamp_db.insert_verification(verification_record)
+            verification_record['block_id'] = None
```
wolfpack94: We set them equal to None because I think the reasoning was that the records can be sent from any of the phases to the phase 5 node and so it doesn't have to be dependent on which phase is sent up to the public broadcast.

blockchain/processing.py
```diff
@@ -606,9 +626,85 @@ def _execute_phase_4(self, config, phase_3_info):
...
+        hashes = []
+        verification_info = {
+            'verification_records': {},
+            'blockchain_type':"BTC"
+        }
+        pending_records = timestamp_db.get_pending_timestamp()
+
+        if not pending_records:
+            return
+
```
wolfpack94: Done.

blockchain/processing.py
```diff
@@ -606,9 +626,85 @@ def _execute_phase_4(self, config, phase_3_info):
...
+        for r in pending_records:
+            hashes.append(r['signature']['hash'])
+            if r['origin_id'] not in verification_info['verification_records']:
+                verification_info['verification_records'][r['origin_id']] = {}
+            verification_info['verification_records'][r['origin_id']][r['timestamp_id']] = r['signature']['hash']
+
+
+        transaction_hash = final_hash(hashes,type=256)
+        verification_info['hash'] = transaction_hash
+
```
wolfpack94: done.

blockchain/processing.py
```diff
@@ -606,9 +626,85 @@ def _execute_phase_4(self, config, phase_3_info):
...
+        stamper = BitcoinTimestamper(self.service_config['bitcoin_network'], BitcoinFeeProvider())
+        bitcoin_tx_id = stamper.persist(transaction_hash)
+        bitcoin_tx_id = b2lx(bitcoin_tx_id).encode('utf-8')
```
wolfpack94: Yes that should be fixed once the new requirements document is installed with pip.

blockchain/processing.py
```diff
@@ -606,9 +626,85 @@ def _execute_phase_4(self, config, phase_3_info):
...
+        verification_info['public_transaction_id'] = bitcoin_tx_id
+
+        prior_block_hash = None
+        lower_hash = None
+        block_id = None
+        phase = 5
+        origin_id = None
+        public_transmission = False
```
wolfpack94: That was my understanding. It came up during one of our code talks in Seattle.

blockchain/processing.py
```diff
@@ -606,9 +626,85 @@ def _execute_phase_4(self, config, phase_3_info):
...
+                                              block_id,
+                                              phase,
+                                              origin_id,
+                                              int(time.time()),
+                                              public_transmission,
+                                              verification_info
+                                              )
+
+        verification_db.insert_verification(block_info['verification_record'])
+
```
wolfpack94: Done.

blockchain/processing.py
```diff
@@ -606,9 +626,85 @@ def _execute_phase_4(self, config, phase_3_info):
...
+        for origin_id in verification_info['verification_records'].keys():
+            for verification_id in verification_info['verification_records'][origin_id]:
+                timestamp_db.set_transaction_timestamp_proof(verification_id)
+
+        unique_vr_transfers = {}
```
wolfpack94: Done.

blockchain/db/postgres/timestamp_db.py
```diff
@@ -0,0 +1,114 @@
...
+    ) VALUES (%s, %s, %s, to_timestamp(%s), %s, %s)"""
+
+
+def get_cursor_name():
+    return str(uuid.uuid4())
+
+
+def set_transaction_timestamp_proof(verification_id):
+    sql_args = (verification_id,)
+    execute_db_args(sql_args, SQL_TIMESTAMP_QUERY)
```
wolfpack94: yes it is the same thing. I will refactor.

blockchain/db/postgres/timestamp_db.py
```diff
@@ -0,0 +1,114 @@
...
+    try:
+        cur = conn.cursor()
+        cur.execute(SQL_INSERT_QUERY, values)
+        conn.commit()
+        cur.close()
+    finally:
+        get_connection_pool().putconn(conn)
+
+
+def get_pending_timestamp():
```
wolfpack94: Done.

blockchain/network.py
```diff
@@ -438,6 +445,9 @@ def phase_1_broadcast(self, block_info, phase_type):
                 message = template.format(type(ex).__name__, ex.args)
                 logger().warning(message)
```
wolfpack94: Might have happened during the merge with master. I deleted the duplicated lines.

blockchain/network.py
```diff
@@ -466,10 +476,25 @@ def phase_3_broadcast(self, block_info, phase_type):
...
-        pass
+        """ send phase_4 information for phase_5 execution """
+        ver_ids = []
+
+        verification_record = block_info['verification_record']
+        verification_info = verification_record['verification_info']
+
+        phase_4_msg = message_types.Phase_4_msg()
+        phase_4_msg.record = thrift_converter.convert_to_thrift_record(verification_record)
+        phase_4_msg.lower_hash = verification_info
```
wolfpack94: This is correct from what we talked about when I was down in Seattle. It is part of the thrift structure.

blockchain/network.py
```diff
@@ -466,10 +476,25 @@ def phase_3_broadcast(self, block_info, phase_type):
...
+
+        for node in self.peer_dict[phase_type]:
```
wolfpack94: This is something that we should be doing. I thought that was a separate issue on here but i may have been mistaken.

blockchain/network.py
```diff
@@ -657,15 +682,16 @@ def public_broadcast(self, block_info, phase):
...
                 verification_record.p3 = phase_msg
             elif phase == 4:
-                pass
+                phase_msg = thrift_converter.get_p4_message(block_info)
+                verification_record.p4 = phase_msg
 
-            phase_5_msg.verification_record = verification_record
+            phase_5_request.verification_record = verification_record
 
             # send block to all known phase 5 nodes
```
wolfpack94: Look at comment above. I can talk again soon.

blockchain/network.py
```diff
@@ -867,4 +899,4 @@ def get_unsent_transfer_ids(self, transfer_to):
 
 if __name__ == '__main__':
     net_dao.reset_data()
-    connect1 = ConnectionManager('localhost', sys.argv[1], sys.argv[3])
+    connect1 = ConnectionManager('localhost', sys.argv[1], sys.argv[3])
```
wolfpack94: Done.

blockchain/network.py
```diff
@@ -806,6 +833,11 @@ def transfer_data(self, pass_phrase, received, unreceived):
                 verifications.append(verification_db.get(guid))
 
             # format verifications to list of thrift structs for returning
+                # retrieve unreceived records
```
wolfpack94: You are correct. It is a duplicate. Fixed.

blockchain/timestamping.py
```diff
@@ -0,0 +1,211 @@
...
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the Apache License with the above modification is
+distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied. See the Apache License for the specific
+language governing permissions and limitations under the Apache License.
+"""
+
+__author__ = "Joe Roets, Brandon Kite, Dylan Yelton, Michael Bachtel"
```
wolfpack94: Thanks for spotting that. Added both.

blockchain/timestamping.py
```diff
@@ -0,0 +1,211 @@
...
+from bitcoin.core import *
+from bitcoin.core.script import *
+
+import requests
+import logging
+
+log = logging.getLogger(__name__)
+
+
+class IPoEStore(object):
```
wolfpack94: Removed.

blockchain/timestamping.py
```diff
@@ -0,0 +1,211 @@
...
+        """
+        Requests a new bitcoin address and returns its public key
+        """
+        proxy = bitcoin.rpc.Proxy()
+        address = proxy.getnewaddress()
+        pubkey = proxy.validateaddress(address)['pubkey']
+        return pubkey
+
+
+    def _build_transaction(self, output, change_pubkey, data):
```
wolfpack94: Refactored the function call. Do you remember what the length allowable with OP_RETURN?

blockchain/transaction_svc.py
```diff
@@ -113,6 +113,13 @@ def set_txn_status(self, log, status_code):
             log.error("500: Internal Error: The server encountered an unexpected condition which prevented it from fulfilling the request.")
         self.set_status(status_code)
 
+# TODO: implement handler for posting transactions to bitcoin network
```
wolfpack94: Just something I forgot to take out when we decided to not implement that in this ticket.

blockchain/timestamping.py
```diff
@@ -0,0 +1,211 @@
...
+        """
+        Builds a bitcoin transaction with 1 input an 2 outputs:
```
wolfpack94: Done.

blockchain/transaction_svc.py
```diff
@@ -113,6 +113,13 @@ def set_txn_status(self, log, status_code):
...
+class public_transmission_handler(tornado.web.RequestHandler):
+    def __init__(self, *args, **kwargs):
+        tornado.web.RequestHandler.__init__(self, *args, **kwargs)
+
+
```
wolfpack94: Removed.

blockchain/transaction_svc.py
```diff
@@ -169,6 +176,8 @@ def run():
     hdlrs = [
         (r"^/transaction$", TransactionHandler),
         (r"^/transaction/(.*)", TransactionHandler),
+        (r"^/public_transmit$", public_transmission_handler),
```
wolfpack94: They were from the implementation of the RestAPI transmission call before we decided to make a separate ticket. Removed.

blockchain/util/crypto.py
```diff
@@ -155,6 +156,8 @@ def sign_verification_record(signatory,
             - arbitrary string
     :return: verification record with assembled signature field
     """
+    # signature, transaction_hash = \
```
wolfpack94: I think that might've been something to do with subscriptions?... I don't recognize it as anything that goes with timestamping. Removed for now but can be added back later if needed.

blockchain/util/crypto.py
```diff
@@ -508,13 +511,18 @@ def deep_hash(thing):
...
 
-def final_hash(items):
+def final_hash(items, type=512):
     """
     Deterministic hashing function that should be used for any full hashing in the system.
     :param items: A list of items to hash deterministically.
     :return: finalized deterministic sha512 hash
     """
-    hash_object = hashlib.sha512(str(deterministic_hash(items)))
+    if type == 512:
```
wolfpack94: Fixed. Thanks for seeing this.

docker/custom-docker-compose.yml
wolfpack94: I do not know what this is... maybe @micylt knows?

thrift/messaging.thrift
wolfpack94: You are correct. Meant to delete. Not used anymore.

docker/custom-docker-compose.yml
Voxelot: this should've been left out by .gitignore, not sure how it got added here.

blockchain/timestamping.py
```diff
@@ -0,0 +1,211 @@
...
+    user name and password are read from ~/.bitcoin/bitcoin.conf configuration file
+
+    usage example:
+        fee_provider = BitcoinFeeProvider()
+        stamper = BitcoinTimestamper("regtest", fee_provider)
+        txid = stamper.persist("hello world!")
+        assert stamper.ispersisted(txid, "hello world!"), "data item not present in the given txid"
+    """
+    MIN_FEE_BYTE = 60
+    MAX_FEE_BYTE = 250
```
Voxelot: Should these be configurable?

blockchain/timestamping.py
```diff
@@ -0,0 +1,211 @@
...
+            A transaction id for referencing the saved data item in the transaction output
+        """
+        proxy = bitcoin.rpc.Proxy()
+        utxo = sorted(proxy.listunspent(0), key=lambda x: hash(x['amount']))[-1]
+        tx = self._build_transaction(utxo, self.get_new_pubkey(), data)
+        value_in = utxo['amount']
+
+        fee_byte = self.fee_provider.recommended()
+        fee_byte = max(fee_byte, self.MIN_FEE_BYTE)
+        assert fee_byte < self.MAX_FEE_BYTE, "too high fee"
```
Voxelot: We should be throwing exceptions instead of using assertions if this is a critical business rule since assertions can be disabled in production.
> Assertions should *not* be used to test for failure cases that can occur because of bad user input or operating system/environment failures, such as a file not being found. Instead, you should raise an exception, or print an error message, or whatever is appropriate. One important reason why assertions should only be used for self-tests of the program is that assertions can be disabled at compile time.
https://wiki.python.org/moin/UsingAssertionsEffectively

blockchain/timestamping.py
```diff
@@ -0,0 +1,211 @@
...
+        r = proxy.signrawtransaction(new_tx)
+        assert r['complete']
+        new_tx = r['tx']
+        suggested_fee = len(new_tx.serialize()) * fee_byte
+        tx_id = proxy.sendrawtransaction(new_tx)
+        log.info('Transaction sent')
+        return tx_id
+
+
+    def get_new_pubkey(self):
```
Voxelot: Does it make sense to be essentially throwing away every token used for timestamping? Is it possible to recombine them over time for future timestamp operations if we keep the private key?

blockchain/processing.py
j0j0r0: Thank you - needed a refresher...
Can you comment per that?

blockchain/processing.py
j0j0r0: Ah, sorry - basically, is there a mechanism to override this (and these) timings?

blockchain/processing.py
j0j0r0: Yes - sorry, I found the answer later in my review - it's used in the query. THX.

blockchain/network.py
j0j0r0: OK - when you're in town, let's look at this with @micylt - need to verify what's in there.

blockchain/network.py
j0j0r0: See below -

blockchain/network.py
j0j0r0: OK - It's already a ticket #87 we will handle there.
Thx.

docker/custom-docker-compose.yml
j0j0r0: OK @wolfpack94 can you remove from the PR?

blockchain/timestamping.py
j0j0r0: Yes - let's @wolfpack94 make a ticket to make these (and maybe other items in the Bitcoin bridge) configuration items.

For example, the fee calculation provider class (I'd like to rename the implemented provider from `BitcoinFeeProvider` to `21coBitcoinFeeProvider).

blockchain/timestamping.py
j0j0r0: @wolfpack94 can you change to throw exceptions herein?

blockchain/timestamping.py
j0j0r0: This may be a question for @lontivero - since it's an `OP_RETURN` transaction, it's a bit different...(see https://21.co/learn/embedding-data-blockchain-op-return/#introducing-op_return)

> What OP_RETURN does is allows an output to contain arbitrary data, simultaneously identifying that it is not a spendable output (not bitcoin being transferred for a later redemption). The opcode says "this is an invalid output" but as long as it is a valid transaction, with an appropriate fee, it will be propagated through the network and mined into a block.

j0j0r0:
@wolfpack94 A few more requests, and we're probably good.

blockchain/timestamping.py
lontivero: @Voxelot i don't understand what you mean by throwing away every token. Could you elaborate that?

As a side comment I see that you are not timestamping a merkle tree root but a hash. Doing this you lose the ability for verifying that a given dragon transaction happened.

Voxelot: @lontivero What I mean by throwing tokens away is that we generate a new address for each timestamp transfer, but don't retain the private key needed to manage any funds transferred to that address. I'm not familiar with how the `OP_RETURN` transaction works so maybe this is not an issue if the wallet is essentially an empty transfer minus the fees.

lontivero: @Voxelot when I designed this, I decided to let all the key management tasks to the Bitcoin Core wallet, just for simplicity. That means that the wallet creates and keeps the key pairs (prv/pub) for us so we don't have to worry about it.

We do retain the private keys. They are keept safe in the wallet.dat file and are available throught the bitcoin rpc wallet methods. However, you say that it could be good to have those prvkey in order to manage funds transferred to that/those address/es but take into account that we are sending *non-spendable outputs* and, for that reason, it doesn't make a lot of sense in our scenarios. *Side comment:* just in case someone sends money to that/those address/es we do have the pk/s to unlock the bitcoins.

Anyway, it is possible to have just one private key (bitcoin address) and buid all the transactions using that as destination, it is not difficult at all. The only drawback is we loose some privacy in the way (an outsider can associate the address with a dragonchain instance and see if the system is running or not, and things like that).

I designed this for trustless verification and with simplicity and privacy in mind. Those principles could be not the best ones. In fact, @j0j0r0 suggested we could use a third-party service for pushing/fetching our txs and in the begining i used a HDW for key management (it is all about tradeoffs)

j0j0r0: Thank you for the extra info @lontivero 

Sounds good.

j0j0r0:
@wolfpack94 Can you add the necessary documentation/comments and verify that everything above is addressed?

I'm available for a call or questions here.

blockchain/processing.py
wolfpack94: Done. Let me know if the comments make sense.

blockchain/processing.py
wolfpack94: There is not at the moment but I can add that to the blockchain4.yml

docker/custom-docker-compose.yml
wolfpack94: Would you like me to remove the "version: '2'" or the whole .gitignore?

blockchain/timestamping.py
wolfpack94: Got it. Now an exception as opposed to an assert.

blockchain/timestamping.py
wolfpack94: Is there still something I need to change?

blockchain/network.py
j0j0r0: This may be the only thing that we need to talk further about still. I'm talking to @micylt tomorrow I think, I'll ask him to review.

docker/custom-docker-compose.yml
j0j0r0: Probably remove the whole file if it's not the single standard one.

blockchain/timestamping.py
j0j0r0: No @wolfpack94 - nothing to change.

blockchain/network.py
j0j0r0: OK - looked at the other phases, and this is the only one that has the `lower_phase` marked in the network module instead of the processing module `execute_phase_x()` function.

It looks like the phase 4 verification info is just a hash (it's only a notary function) but I want to know why this is set within the network module unlike the other phases.

If you have time this week or weekend, can we do a quick walkthrough?

blockchain/processing.py
j0j0r0: Looks good. Thanks.

blockchain/processing.py
j0j0r0: Or you can add a ticket for that (probably preferred).

blockchain/processing.py
j0j0r0: Yes - IIRC, also that we may be placing multiple blocks (with different block IDs) from multiple nodes into this phase 5 record.

Can you comment to that effect for each of these lines? TODO

blockchain/processing.py
```diff
@@ -676,6 +676,9 @@ def _execute_timestamping(self, config):
         # takes the list of hashes to be transmitted and hashes with 256 bit to get in form to send
         transaction_hash = final_hash(hashes, type=256)
 
+        # normal SHA512 hash for all lower VR contents
+        lower_hash = final_hash(hashes)
```
j0j0r0: FYI  - added this to do the consistent SHA512 hashing on all lower VR contents of this phase 5 VR.
:)

Issue #99, opened 2017-03-08
Smart Contract Expansion


j0j0r0:
Expand the smart contract (SC) system.

Implementation of the following: 

1. Design for the creation of "reserved" SC - e.g. subscription provisioning, smart contract provisioning, currency operations, marketplace operations, etc. (partially implemented)
1. Design for the structure of the SC dictionary(s) and ability to replace and override
1. Transaction event SC - for approval/denial of individual transactions from within `Transaction Service`
1. Subscription feed based SC - executed against matching _transactions and verification records_ received via subscription.
1. Arbitrary or Library SC - Custom utility code or library code that may be executable from within other SC or by the system at large - e.g. shared user code, currency or marketplace functionality, etc.

*Language extensions should not be considered within this ticket - another ticket shall be created*

j0j0r0:
Currently, I _think_ that the SC may be best separated into separate dictionaries/structures. _Quick thoughts..._

- It _may_ be prudent to separate the reserved SC from the transaction SC so that user configuration would not overwrite the original system function
- The transaction event SC are triggered based solely upon the transaction type, and thus should _probably_ be separated from the subscription SC as they may have duplicate transaction type (not unique in dictionary)
- Subscription SC may have more complex definition to enable a selection criteria (e.g. TT_X + VR3)
- Arbitrary/Library SC will probably have a simple name lookup, but separation

j0j0r0:
Types of SC: 

1. Transaction SC (TSC)
1. Subscription SC (SSC)
1. Library SC (LSC)
1. BroadcastReceipt SC (BSC)

TSC will be divided on implementation to be held in 2 separate structures, one for user/business transactions, and another for reserved transactions.

TSC,  LSC, and BSC will use a simple string key for lookup.
SSC will use composite key of transaction type/name, VR phase, and origin node ID for lookup.

j0j0r0:
We probably need to abstract the divisions between the SC and allow a simple interface.

j0j0r0:
### I/O necessary from within SC

#### TSC

- Transaction in, approve/deny response
- Read only access to blockchain DB table subset

#### SSC

- Access to application DB? (not dragonchain/blockchain DB)
- Network Access?

#### LSC

- Network access?

#### BSC

- Access to application DB? (not dragonchain/blockchain DB)

j0j0r0:
[...]

PR #108, opened 2017-04-03
Smart Contracts - Trusted

micylt:
issue #99 

The features added in this pull request give users the ability to post smart contract transactions.

Posting a smart contract transaction triggers a provisioning phase.
   * smart contract stored in code.
   * smart contract info stored in database.

Type of smart contracts:
* Transaction SC (TSC)
   * divided to be held in 2 separate structures, one for user/business transactions, and another for reserved transactions
   * stored in dictionary
      * key: transaction type
      * value: sc function
   * transactions posted with transaction type will run smart contract code if applicable.
* Subscription SC (SSC)
   * smart contract run on subscription transactions + verifications received.
   * stored in dictionary
      * key: origin_id:txn_type:phase (one or more)
      * value: sc function
* Library SC (LSC)
   * currency
* BroadcastReceipt SC (BSC)
   * smart contract run on verifications received from broadcast receipts.
   * stored in dictionary
      * key: phase
      * value: sc function

* Smart contract data stored:
  * sc_id: UUID
  * sc_class: TSC, SSC, etc.
  * smart_contract: sc code
  * sc_key: transaction type, phase, etc.
  * criteria: transaction type, phase, etc. (what must be included in payload for sc execution)
  * test: unit test(s)
  * requirements: required libraries, etc.
  * version: sc version num.
  * status: active, pending, etc.

* Any existing smart contracts are loaded from database at startup and are stored into their appropriate structures.

* example of a valid BSC Provisioning transaction:
```
{
  "header": {
    "create_ts": 1475180987,
    "business_unit": "a3e13076-8683-11e6-97a9-3c970e3bee11",
    "family_of_business": "Test Business Family",
    "line_of_business": "My Business",
    "owner": "Test Node",
    "transaction_type": "TT_PROVISION_BSC",
    "actor": "c26dd972-8683-11e6-977b-3c970e3bee11",
    "entity": "c78f4526-8683-11e6-b1c6-3c970e3bee11"
  },
  "payload": {
    "phase": 2,
    "smart_contract": {
      "bsc": "def func(self, transaction): print transaction['header']['transaction_type']"
    },
    "criteria": ["phase"],
    "test": "print 'TEST'",
    "requirements": ["uuid", "time"],
    "version": 1
  },
    "source": "f36c9086-8683-11e6-80dc-3c970e3bee11"
}
```

currently for smart contract code to work the code provided must start with def func(...):

This implementation runs under the assumption that the user is posting valid unmalicious smart contract code.

An untrusted version will be implemented in the future, which will contain smart contract logic in jailed root directories to prevent users from accessing restricted data, network capabilities, etc.


blockchain/smart_contracts/sc_trusted.py
```diff
@@ -0,0 +1,229 @@
...
+        try:
+            sc_dao.insert_sc(pl, sc_class, sc_key)
+        except Exception as ex:
+            template = "An exception of type {0} occurred. Arguments:\n{1!r}"
+            message = template.format(type(ex).__name__, ex.args)
+            logger().warning(message)
+            return False
+        return True
+
+    # FIXME: only load highest version sc to avoid overwriting
```
j0j0r0: Is this a ticket, or should it be completed here?

blockchain/smart_contracts/sc_trusted.py
```diff
@@ -0,0 +1,229 @@
...
+                    # define sc function
+                    exec(sc['smart_contract'])
+                    self.sc_container[sc_class][sc_key] = func
+        except Exception as ex:
+            template = "An exception of type {0} occurred. Arguments:\n{1!r}"
+            message = template.format(type(ex).__name__, ex.args)
+            logger().warning(message)
+
+
+if __name__ == '__main__':
```
j0j0r0: Was this here for test/dev?

blockchain/network.py
```diff
@@ -556,12 +562,39 @@ def insert_verifications(self, verification_records):
...
+        """ execute subscription smart contract """
+        for sc_key in self.processing_node.scp.ssc.keys():
+            # [origin_id:txn_type:phase]
+            criteria = sc_key.split(":")
+            origin_id = criteria[0]
+            txn_type = criteria[1]
+            phase = criteria[2]
+            vrs = verification_db.get_all(origin_id=origin_id, phase=phase, min_block_id=min_block_id)
+            for v in vrs:
+                txns = transaction_db.get_all(txn_type, v['block_id'])
```
Voxelot: This could lead to a lot of redundant and expensive transaction queries (e.g. 5 p2 vr's for the same set of txns). Maybe create a dictionary of (txn_type, blockid) tuples mapped to lists of vrs first, then only fetch txns by unique combinations of txn_type and blockid, then invoke the ssc for each vr associated with that tuple.

blockchain/smart_contracts/sc_trusted.py
```diff
@@ -0,0 +1,229 @@
...
+            else:
+                return False
+        except Exception as ex:
+            template = "An exception of type {0} occurred. Arguments:\n{1!r}"
+            message = template.format(type(ex).__name__, ex.args)
+            logger().warning(message)
+            return False
+        return True
+
+    def _insert_sc(self, pl, sc_class, sc_key):
```
Voxelot: Think we might want a way to unprovision a SC in the future? If so that api should also be made available to the SC (although it shouldn't have access to other SC's). That way it could de-register itself once some terms or conditions have been satisfied. One example could be some sort of bond-like instrument that pays out from a wallet on a fixed interval and then unprovisions itself once cashed out.

scripts/valid_payload.json
```diff
@@ -5,27 +5,19 @@
     "family_of_business": "Test Business Family",
```
Voxelot: since this is kind of used as an example, maybe we should start breaking these out into separate files so we have some sort of reference library of core transaction schema examples?

blockchain/smart_contracts/sc_trusted.py
```diff
@@ -0,0 +1,229 @@
...
+    return logging.getLogger(name)
+
+
+class SmartContractProvisioning(object):
+    def __init__(self, network=None, public_key=None):
+        # processing nodes network
+        self.network = network
+        # processing node's public key
+        self.public_key = public_key
+        self.rtsc = {"TT_SUB_REQ": self.subscription_request,
```
Voxelot: agreed, I think we should confine the definition of smart contract to custom business rules submitted via transactions that aren't already built into the platform.

blockchain/db/postgres/smart_contracts_db.py
```diff
@@ -0,0 +1,109 @@
...
+from postgres import get_connection_pool
+import uuid
+
+from blockchain.qry import format_sc
+
+
+""" CONSTANTS """
+DEFAULT_PAGE_SIZE = 1000
+""" SQL Queries """
+SQL_GET_ALL = """SELECT * FROM smart_contracts"""
```
micylt: I concatenate the WHERE here...

https://github.com/dragonchain/dragonchain/pull/108/files#diff-88b39a52f3a047b4d8488bb0aaceda9fR94

Would you prefer we refactor the constants to include WHERE?

blockchain/db/postgres/transaction_db.py
```diff
@@ -89,61 +89,70 @@ def get_all(limit=None, offset=None, **params):
         query += """ WHERE """
 
     if "block_id" in params:
-        query += """ block_id = """ + str(params["block_id"])
-        separator_needed = True
+        if params["block_id"]:
```
micylt: Done. Refactored both.

blockchain/network.py
```diff
@@ -556,12 +562,39 @@ def insert_verifications(self, verification_records):
                 logger().error(message)
                 continue
 
+    def get_min_block_id(self, vrs):
```
micylt: Yes, that works! Much cleaner. Thanks for pointing that out. Done.

blockchain/network.py
```diff
@@ -556,12 +562,39 @@ def insert_verifications(self, verification_records):
...
+                id = vrs[i]['block_id']
+                if id < min_id:
+                    min_id = id
+        return min_id
+
+    def _execute_ssc(self, min_block_id):
+        """ execute subscription smart contract """
+        for sc_key in self.processing_node.scp.ssc.keys():
+            # [origin_id:txn_type:phase]
+            criteria = sc_key.split(":")
```
micylt: Done.

blockchain/network.py
```diff
@@ -556,12 +562,39 @@ def insert_verifications(self, verification_records):
...
+                txns = transaction_db.get_all(txn_type, v['block_id'])
+                self.processing_node.scp.ssc[sc_key](txns, v)
+
     def resolve_data(self, verifications, phase):
         """ store received verifications from node, find replications and store them in transfers table """
         for verification in verifications:
             try:
                 verification = thrift_converter.convert_thrift_verification(verification)
                 if verification['signature']['signatory'] is not self.this_node.node_id:
+                    # run smart contract code if phase matches
```
micylt: Yes. Made that more clear in the above comment.

blockchain/smart_contracts/sc_trusted.py
micylt: Yes. I added this to run on its own for testing. I will take it out.

blockchain/db/postgres/smart_contracts_db.py
j0j0r0: I've modified to move the `WHERE` into `GET_APPROVED` constant.

blockchain/network.py
j0j0r0: +1

blockchain/smart_contracts/sc_trusted.py
j0j0r0: OK - let's talk about this this week (4/10/2017).

We may need to have another parameter to point to the SC implementation/module as well.

blockchain/smart_contracts/sc_trusted.py
j0j0r0: I've created ticket #116.

blockchain/processing.py
```diff
@@ -218,44 +219,13 @@ def get_prior_hash(self, origin_id, phase):
 
         return prior_hash
 
-    def _subscription_request(self, transaction):
```
j0j0r0: Same above - speak this week to resolve.

blockchain/processing.py
```diff
@@ -340,6 +306,34 @@ def _execute_phase_1(self, config, current_block_id):
...
+            check if the given transaction falls under one of these smart contract types and execute
+            the appropriate smart contract provisioning function on it
+         """
+        txn_type = txn["header"]["transaction_type"]
+        # reserved transaction smart contracts
+        rtsc = self.scp.rtsc
+        if txn_type in rtsc:
+            if rtsc[txn["header"]["transaction_type"]](txn):
+                return True
+        elif txn_type == "TT_PROVISION_TSC":
```
j0j0r0: I'm thinking that we'll want the caller to specify the SC implementation, with TT naming being same for all (per SC type).

blockchain/smart_contracts/sc_trusted.py
j0j0r0: Interesting idea.
And, yes, we'll need to be able to manually specify that a SC is no longer active, and let the SC be able to do that as well.

We _could_ (should?) make it so that some SC could own other SC and could either completely control others or themselves, or in combination in multi-sig configuration with other SC or humans.

scripts/valid_payload.json
j0j0r0: Yes, we should have one or more examples for any given transaction type (esp. like these SC provisioning transactions) and we can use them for documentation and even testing.

When we finalize, we should create a ticket for that.

blockchain/smart_contracts/sc_trusted.py
j0j0r0: Note:
To create:

- `smart_contracts.py`
- `reserved_sc.py`

Move all reserved SC into `reserved_sc.py`...
`smart_contracts.py` will contain generic provisioning and calling code.

blockchain/smart_contracts/sc_trusted.py
j0j0r0: Issue #118

scripts/valid_payload.json
j0j0r0: #119

blockchain/smart_contracts/sc_trusted.py
j0j0r0: TODO move all calls and provisioning to call `smart_contracts.py` module to abstract all calls.

blockchain/network.py
Voxelot: Also... This may return transactions not included in the verification record if the TT is the same across several subscriptions. We could include owner/origin in the query (e.g. origin - original phase_1 node_id). We also need pagination here if there are more than 10 transactions.

Voxelot: something like this...

```
queries = dict()
for vr in vrs:
    if (vr['block_id'], vr['origin_id'], ) in queries:
        queries[(vr['block_id'], vr['origin_id'], )] += [vr]
    else:
        queries[(vr['block_id'], vr['origin_id'], )] = [vr]

for query in queries:
    txns = transaction_db.get_all(limit = (1<<31)-1, block_id=query[0], owner=query[1])
    for vr in queries[query]:
        self.processing_node.scp.ssc[sc_key](txns, vr)
```

blockchain/smart_contracts/smart_contracts.py
```diff
@@ -0,0 +1,268 @@
...
+            return False
+        return True
+
+    def execute_ssc(self, min_block_id):
+        """ execute subscription smart contract """
+        for sc_key in self.ssc.keys():
+            (origin_id, txn_type, phase) = sc_key.split(":")
+            vrs = verification_db.get_all(origin_id=origin_id, phase=phase, min_block_id=min_block_id)
+            for v in vrs:
+                txns = transaction_db.get_all(txn_type, v['block_id'])
```
Voxelot: You added the origin, but you still didn't address the redundancy issue...

Here's an example scenario:
I subscribe to a server and require 10 p2 verifications per block. During a 5 second window, the server I'm subscribed to processes 100 txns that match the subscription. Using this approach, you'll end up fetching the same set of 100 txns 10 times (10 vrs). If you know that the vrs are all grouped with the same set of transactions (block + origin), you could get away with only fetching 100 instead of 1000 txns by reusing them for multiple smart contract executions.

j0j0r0: @micylt We can review this and talk early this week if you're available.
Also, can you resolve the conflicts of this branch listed at bottom?

j0j0r0:
TODO review @Voxelot comment and resolve conflicts.

blockchain/smart_contracts/smart_contracts.py
Voxelot: Oh yeah, this also doesn't override the default limit of 10. So continuing my example, the SC would only execute with 10% of the txns it was expecting to see.

blockchain/processing.py
j0j0r0: Completed to satisfaction.

Issue #130, opened 2017-08-07
Base64 Encode Smart Contracts


Voxelot:
In order to avoid escaping issues, smart contract scripts should be base64'd in the transaction payload and decoded before execution setup.

cheeseandcereal:
stale

PR #131, opened 2017-08-07
#130 base64 decode smart contracts

Voxelot:
Allowing smart contracts to be base64 encoded to avoid escaping issues etc...


j0j0r0:
If you disagree, let me know and  I'll merge as is. 
Also, if we do fail non encoded, we should tag a new version.

Voxelot:
Since the feature is still in progress I suppose we can give up backwards compatibility here to keep things streamlined. I'll update the smart contract docs.

PR #134, opened 2017-10-02
Issue 37/refactoring lower phase hash

ghost:


thrift/messaging.thrift
```diff
@@ -136,10 +136,10 @@ struct Phase_2_msg {
 /* TODO: rename business_list and deploy_loc_list to businesses and deploy_locations respectively */
```
Voxelot: Looks like we can get rid of the todo now too! Thanks :)

cheeseandcereal:
stale/outdated

PR #141, opened 2017-11-19
Set up more automation and install scripts ;)

TheRoboKitten:
Hell, I give myself an a++ for doing this.

Simpy run the install script and you are good to go given you update the config template with already-known nodes.

Check it out 🔢 🥇 💯 👍 

The ONLY thing that needs to be revised is "TransactionTSC.json" in scripts to match SSC transaction provision. Have fun.


TheRoboKitten:
My only issue is docker does not like to start for some reason on install script :( so I docker up twice, and that still doesn't give the script time to pull in the nodes. So it has to be done manually for each port.

SOMETIMES it works. but 50% of the time it doesn't.

PR #143, opened 2017-11-20
AES CFB Over-The-URL encryption on POST.

Gr1dd:
My next step will be completely encrypting the GET/HTTP requests for better API protection using the same Private key. Dunno. Mainly PRing this branch to check for differences, let me know if any code updates need to happen.

CFB allows for multiple lengths of json data with padding and un-padding the data.
Should also allows for multiple varying different key sizes. (atleast 64-bit key for AES256 is required I believe)
So you'd need 2x SHA256 hash lengths.

I removed sub request approval to auto-approve (needs testing) and should not be needed in local networks or public networks as long as everyone shares the same key and are heavily guarded.

### DIFF
--- a/docs/install.txt
+++ b/docs/install.txt
@@ -38,9 +38,9 @@
 # ...OR you could subscribe to the node's transactions:
 # Edit the subscription0.json or subscription1.json in scripts/
 # sudo nano subscription0.json
-# INPUT the phase requirement, phase, host, and node ID then: ctrl + o to save ctrl + x to exit then:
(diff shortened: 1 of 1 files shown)


Gr1dd:
Going to squash my commits.

Voxelot:
Hey love seeing the work you guys are doing! Before you submit the next PR please think about how you'd break these changes up a bit. For instance, if you're playing with dockerization please keep that separate from changes to the REST APIs. Ideally, each PR shouldn't touch more than 2-4 files.

PR #145, opened 2017-11-21
Updates/PR for comparison

Gr1dd:
Lots of stuff because gen_thrift had to be run.
The main points:
* Added AES encryption on POST requests through url-safe encoding.
* Added bogus verification in thrift file so thrift would generate under phase 5 (because it was undefined)
* Added Install Script
* Added start/stop scripts
* Left my default key in-place, removed key generation on install script.
* Added python script to send a transaction json
* Added configuration template that is auto-generated for localhost with a fresh UUID upon running install.sh
* Edited sub_to_db.py to insert subscriptions as approved because I could not figure out how to approve node subscriptions:
```
-        "pending"  # subscription status
 +        "approved"  # subscription status
      )
```
* Turned off phase broadcast to phase 5 by default in config template preventing a plethora of errors, and promoting quicker deployment time.

* Set a install.txt note in the scripts.


Gr1dd:
Fix your software.

PR #157, opened 2018-01-03
Resolved: Invalid function close_connection_pool

Rhett-Trappman:
postgres.close_connection_pool() doesn't exist, should use postgres.cleanup().


cheeseandcereal:
stale

PR #186, opened 2019-09-05
Support changes to matchmaking error code updates

rdaquilante:
- [x] tests are included
- [ ] documentation is changed or added


cheeseandcereal:
This has to be released with https://github.com/dragonchain-inc/matchmaking/pull/183 in both staging and production. So we'll wait to merge for now until we can release them both through the whole pipeline at once.

cheeseandcereal:
Updated and tested working in staging.

PR #190, opened 2019-09-16
update redisearch error check

cheeseandcereal:
## Description

Redisearch changed their error strings which require this update

##### Checklist

### DIFF
--- a/dragonchain/lib/database/redisearch.py
+++ b/dragonchain/lib/database/redisearch.py
@@ -186,7 +186,7 @@ def delete_index(index: str) -> None:
     try:
         client.drop_index()
     except redis.exceptions.ResponseError as e:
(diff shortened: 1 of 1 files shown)


dragonchain/lib/database/redisearch.py
```diff
@@ -186,7 +186,7 @@ def delete_index(index: str) -> None:
     try:
         client.drop_index()
     except redis.exceptions.ResponseError as e:
-        if str(e) != "Unknown Index name":  # Don't care if the index doesn't exist when trying to delete
+        if not str(e).startswith("Unknown Index name"):  # Don't care if the index doesn't exist when trying to delete
```
deanshelton913: man this sucks that we cant use a code or something

cheeseandcereal: I agree, but unfortunately the redisearch package doesn't handle errors at all, and just passes back the raw redis string.

PR #191, opened 2019-09-16
OpenFaaS logs endpoint

regan-karlewicz:
## Description

Adds new endpoint to the Dragonchain webserver that exposes the OpenFaaS log endpoint for smart contracts. The endpoint is in essence a pass through to the OpenFaaS cluster, and returns a list of logging events. The endpoint is as such:

```
GET /v1/contract/logs/<contract_id>?tail=<maximum # logs to return>&since=<RFC3339 timestamp>
```

and returns

```
{
    "logs": [{"name": string, "instance": string, "timestamp": string, "text": string}, ...]
}
```

##### Checklist

- [X] `./tools.sh full-test` passes
- [x] tests are included


dragonchain/job_processor/contract_job.py
```diff
@@ -34,6 +34,7 @@
 from dragonchain.lib import keys
 from dragonchain.lib.interfaces import storage
 from dragonchain.lib.interfaces import registry as registry_interface  # Alternate naming to alleviate confusion
+from dragonchain.lib.faas import get_faas_auth
```
cheeseandcereal: only import modules, not functions

dragonchain/lib/logs.py
```diff
@@ -0,0 +1,26 @@
+import os
```
cheeseandcereal: why is this file called logs.py when it's very specifically for openfaas? Should probably go in faas.py

dragonchain/lib/logs.py
```diff
@@ -0,0 +1,26 @@
...
+import json
+from typing import List, Dict, Optional, cast, Any
+
+import requests
+
+from dragonchain.lib.faas import get_faas_auth
```
cheeseandcereal: don't import functions directly

dragonchain/lib/logs.py
```diff
@@ -0,0 +1,26 @@
...
+    if response.status_code != 200:
+        raise RuntimeError("Error getting contract logs, non-2XX response from OpenFaaS gateway")
+
+    return response.text.split("\n")
+
+
+def get_logs(contract_id: str, since: Optional[str] = None, tail: Optional[int] = 100) -> List[Dict[str, str]]:
+    """Gets the raw logs from openfaas and parses the ndjson into a list of dictionaries"""
+    raw_logs = get_raw_logs(contract_id, since, tail)
+    return list(map(lambda x: json.parse(x), raw_logs))
```
cheeseandcereal: this is not very pythonic. use list comprehension instead. ie `return [json.parse(x) for x in raw_logs]`

dragonchain/lib/logs.py
```diff
@@ -0,0 +1,26 @@
...
+FAAS_GATEWAY = os.environ["FAAS_GATEWAY"]
+
+
+def get_raw_logs(contract_id: str, since: Optional[str] = None, tail: Optional[int] = 100) -> List[str]:
+    """Calls openfaas /system/logs endpoint with query parameters for a specific contract"""
+    endpoint = f"{FAAS_GATEWAY}/system/logs"
+    query_params = cast(Dict[str, Any], {"name": f"contract-{contract_id}", "tail": tail, "since": since})
+    response = requests.get(endpoint, params=query_params, headers={"Authorization": get_faas_auth()})
+    if response.status_code != 200:
+        raise RuntimeError("Error getting contract logs, non-2XX response from OpenFaaS gateway")
```
cheeseandcereal: We should probably consider having a faas error or something so that's it's easily catchable if we try to catch in the future

dragonchain/webserver/routes/smart_contracts.py
```diff
@@ -35,6 +35,8 @@ def apply_routes(app: flask.Flask):
     app.add_url_rule("/v1/contract", "list_contract_v1", list_contract_v1, methods=["GET"])
     app.add_url_rule("/contract/<contract_id>", "get_contract_by_id_v1", get_contract_by_id_v1, methods=["GET"])
     app.add_url_rule("/v1/contract/<contract_id>", "get_contract_by_id_v1", get_contract_by_id_v1, methods=["GET"])
+    app.add_url_rule("/contract/logs/<contract_id>", "get_smart_contract_logs_v1", get_smart_contract_logs_v1, methods=["GET"])
```
cheeseandcereal: don't need the unversioned route for new endpoints. Only reason those exist on other routes is for legacy support.

dragonchain/lib/faas.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    with open("/etc/openfaas-secret/user", "r") as file:
+        username = file.read()
+    with open("/etc/openfaas-secret/password", "r") as file:
+        password = file.read()
+
+    return f"Basic {base64.b64encode(f'{username}:{password}'.encode('utf-8')).decode('ascii')}"
+
+
+def get_raw_logs(contract_id: str, since: Optional[str] = None, tail: Optional[int] = 100) -> List[str]:
```
cheeseandcereal: is there a reason to have a `get_logs` as well as a `get_raw_logs`? It looks like you're only using `get_logs`. Could be confusing to keep both and know which one to use in the future. Is there any reason why we'll ever want to use `get_raw_logs` separately?

regan-karlewicz: I just did it to separate the call from the transforming logic. I'll mark it as private with the leading `_`

cheeseandcereal:
Before merging this, we should also get SDK PRs staged, and integration tests added (with the python sdk)

dragonchain/lib/faas.py
```diff
@@ -0,0 +1,45 @@
...
+    with open("/etc/openfaas-secret/password", "r") as file:
+        password = file.read()
+
+    return f"Basic {base64.b64encode(f'{username}:{password}'.encode('utf-8')).decode('ascii')}"
+
+
+def _get_raw_logs(contract_id: str, since: Optional[str] = None, tail: Optional[int] = 100) -> List[str]:
+    """Calls openfaas /system/logs endpoint with query parameters for a specific contract"""
+    query_params = cast(Dict[str, Any], {"name": f"contract-{contract_id}", "tail": tail, "since": since})
+    headers = {} if os.environ.get("FAAS_LOGS_GATEWAY") else {"Authorization": faas.get_faas_auth()}
```
cheeseandcereal: my vote is have `get_faas_auth` simply return an empty string if it errors, rather than throwing an exception, then always include this header. Cleans up this potentially confusing branch which isn't obvious it's for a proxy

dragonchain/lib/faas.py
```diff
@@ -0,0 +1,45 @@
+import os
+import base64
+import json
+from typing import List, Dict, Optional, cast, Any
+
+import requests
+
+from dragonchain.lib import faas
```
cheeseandcereal: why is it importing itself?

regan-karlewicz: Good question Lol

.version
```diff
@@ -1 +1 @@
-4.0.0
\ No newline at end of file
+4.0.0-c
```
cheeseandcereal: can't approve merge until this is reverted

dragonchain/lib/faas.py
```diff
@@ -0,0 +1,46 @@
...
+
+from dragonchain import exceptions
+
+FAAS_GATEWAY = os.environ.get("FAAS_LOGS_GATEWAY") or os.environ["FAAS_GATEWAY"]
+
+
+def get_faas_auth() -> str:
+    """Gets authorization to use OpenFaaS
+
+        Returns:
```
cheeseandcereal: Un-indent for correct docstring and is probably worth mentioning in the docstring that this returns an empty string if it can't find the openfaas secrets rather than throwing an exception

dragonchain/lib/faas.py
```diff
@@ -0,0 +1,46 @@
...
+            raise exceptions.NotFound("Logs not found for contract")
+        else:
+            raise exceptions.OpenFaasException("Error getting contract logs, non-2XX response from OpenFaaS gateway")
+
+    return response.text.split("\n")
+
+
+def get_logs(contract_id: str, since: Optional[str] = None, tail: Optional[int] = 100) -> List[Dict[str, str]]:
+    """Gets the raw logs from openfaas and parses the ndjson into a list of dictionaries"""
+    raw_logs = list(filter(lambda x: x != "", _get_raw_logs(contract_id, since, tail)))
```
cheeseandcereal: you removed one ugly lambda and replaced it with another lol. I would recommend list comprehension here too for readability

`[x for x in _get_raw_logs(contract_id, since, tail) if x != ""]`

dragonchain/webserver/routes/smart_contracts.py
```diff
@@ -64,6 +65,13 @@ def list_contract_v1() -> Tuple[str, int, Dict[str, str]]:
     return helpers.flask_http_response(200, smart_contracts.list_contracts_v1())
 
 
+@request_authorizer.Authenticated()
+def get_smart_contract_logs_v1(contract_id: str) -> Tuple[str, int, Dict[str, str]]:
+    since = flask.request.args.get("since")
+    tail = flask.request.args.get("tail")
```
cheeseandcereal: This will always be a string. All request args are always strings. You'll want to parse this into an int and throw a reasonable error (400) if it's not parsable into an int.

I'm surprised mypy didn't pick this up. I guess flask isn't typed or something.

cheeseandcereal: Demo:
```python
import flask
app = flask.Flask(__name__)

@app.route("/")
def test():
    return str(type(flask.request.args.get("test")))

app.run()
```

```sh
$ curl localhost:5000/?test=123
<class 'str'>
```

dragonchain/lib/faas.py
```diff
@@ -0,0 +1,54 @@
...
+
+FAAS_GATEWAY = os.environ.get("FAAS_LOGS_GATEWAY") or os.environ["FAAS_GATEWAY"]
+
+
+def get_faas_auth() -> str:
+    """Gets authorization to use OpenFaaS
+
+    Returns:
+        A string containing authorization for OpenFaaS.
+        Returns empty string if authorization is not found, in the case of managed services
```
cheeseandcereal: I wouldn't say "in the case of managed services"

I would just say something like "if openfaas secrets aren't configured"

PR #193, opened 2019-09-18
Adds verification notification system

deanshelton913:
# Verification Notification

## Use Case
How can I know when verifications come-in for my blocks in an event-based way?

## Proposal:
Provide configuration option which allows Dragonchain to transmit via HTTP, all received verification objects. 

## Description
This PR modifies the receipt endpoint on Dragonchain to re-transmit any received verifications to a unique list of preconfigured URLs. 

##### Checklist

- [x] tests are included


.version
```diff
@@ -1 +1 @@
-4.0.0
\ No newline at end of file
+4.1.0
```
cheeseandcereal: Not released yet so this wouldn't need a version bump

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## 4.1.0
```
cheeseandcereal: No version bump required since everything in 4.0.0 is unreleased. Just move down into 4.0.0

dragonchain/webserver/lib/dragonnet.py
```diff
@@ -80,13 +87,51 @@ def process_receipt_v1(block_dto: Dict[str, Any]) -> None:
                 _log.exception("matchmaking add_receipt failed!")
             # Update the broadcast system about this receipt
             broadcast_functions.set_receieved_verification_for_block_from_chain_sync(l1_block_id, level_received_from, block_model.dc_id)
+            # Send verifications to any configured urls
+            send_verification_notifications(level_received_from, block_model)
```
cheeseandcereal: Not sure if this is the best place to put this. This means that any time a dragon net node sends a receipt to an L1, that http request is now hung up on that L1 making a POST to another unrelated server. I can forsee potential timeout issues here. Blocking a webserver request on another arbitrary POST seems like a potentially bad idea

docs/static/chart/opensource-config.yaml
```diff
@@ -96,6 +96,34 @@ dragonchain:
         requests:
           storage: 2Gi
 
+
+  # -- dragonchain.verificationNotifications --
+  # This Level 1 Dragonchain configuration will send received verifications via HTTP to up to 10 unique URL's per level configured here.
```
regan-karlewicz: Any reason for the limitation of 10? If the user wants to send to 1000 endpoints, and it slows down their chain, shouldn't that be on them?

docs/static/chart/opensource-config.yaml
```diff
@@ -96,6 +96,34 @@ dragonchain:
...
+  # Verifications will be sent with HTTP POST, and contain three relevant headers documented below.
+  # -- Headers --
+  # Authentication: Bearer ${signature}
+  # DragonchainId: PublicIdOfThisL1Dragonchain
+  # Content-Type: application/json
+  # -- Body --
+  # The postbody will be JSON of the raw verification object received by this dragonchain. See Dragonchain documentation for details.
```
regan-karlewicz: `post body`

docs/static/chart/opensource-config.yaml
```diff
@@ -96,6 +96,34 @@ dragonchain:
...
+  # -- Signature --
+  # The bearer token will be the signature of the utf8 encoded bytes of the postbody signed with the private key, and verifyable with
```
regan-karlewicz: `post body`, `verifiable`

dragonchain/webserver/lib/dragonnet.py
```diff
@@ -80,13 +87,51 @@ def process_receipt_v1(block_dto: Dict[str, Any]) -> None:
...
+        urls = VERIFICATION_NOTIFICATION[key]
+    except KeyError:
+        pass
+    if urls is None:
+        urls = []
+    return set(urls)
+
+
+def sign(message: str) -> Tuple[str, Optional[str]]:
+    dc_keys = keys.DCKeys()
```
cheeseandcereal: use `keys.get_my_keys()` instead of `keys.DCKeys()`

dragonchain/webserver/lib/dragonnet.py
```diff
@@ -80,13 +87,51 @@ def process_receipt_v1(block_dto: Dict[str, Any]) -> None:
...
 
 
+def get_notification_urls(key: str) -> set:
+    urls = None
+    try:
+        urls = VERIFICATION_NOTIFICATION[key]
+    except KeyError:
+        pass
+    if urls is None:
+        urls = []
```
regan-karlewicz: Shouldn't this return `set()`?

dragonchain/webserver/lib/dragonnet.py
```diff
@@ -80,13 +87,51 @@ def process_receipt_v1(block_dto: Dict[str, Any]) -> None:
...
             )
             raise exceptions.NotAcceptingVerifications(f"Not accepting verifications for block {l1_block_id} from {block_model.dc_id}")
 
 
+def get_notification_urls(key: str) -> set:
+    urls = None
+    try:
+        urls = VERIFICATION_NOTIFICATION[key]
+    except KeyError:
+        pass
```
regan-karlewicz: Could change this to `urls = set()` and then you can get rid of line 100, 105, 106

dragonchain/webserver/lib/dragonnet.py
```diff
@@ -80,13 +87,51 @@ def process_receipt_v1(block_dto: Dict[str, Any]) -> None:
...
+    if urls is None:
+        urls = []
+    return set(urls)
+
+
+def sign(message: str) -> Tuple[str, Optional[str]]:
+    dc_keys = keys.DCKeys()
+    signature = dc_keys.make_signature(message.encode("utf8"), crypto.SupportedHashes.sha256)
+    return (signature, dc_keys.pub)
```
cheeseandcereal: `dc_keys.pub` is an internal C secp256k1 struct. The chain's own public id is it's base58 encoded public key. I would recommend we use that instead.

It also seems unnecessary to return this for every signature, as a chain's public key won't change while it's running. Whatever function needing to use the public key string can call `keys.get_public_id()` when it needs it rather than passing this static key through a call stack every time

dragonchain/webserver/lib/dragonnet.py
```diff
@@ -80,13 +87,51 @@ def process_receipt_v1(block_dto: Dict[str, Any]) -> None:
...
+
+
+def send_verification_notifications(level: int, block_model: "model.BlockModel"):
+    try:
+        if VERIFICATION_NOTIFICATION:
+            string_json = json.dumps(block_model.export_as_at_rest())
```
cheeseandcereal: add separators tuple as per every other json.dumps in the project

dragonchain/webserver/lib/dragonnet.py
```diff
@@ -80,13 +87,48 @@ def process_receipt_v1(block_dto: Dict[str, Any]) -> None:
...
+            (signature, public_key) = sign(string_json)
+            all_endpoints = get_notification_urls("all").union(get_notification_urls(f"l{level}"))
+            for url in all_endpoints:
+                _log.debug(f"Notify -> POST {url} {string_json}")
+                response = requests.post(
```
cheeseandcereal: never ever do an http request without a timeout. This can stall the pod permanently

docs/static/chart/opensource-config.yaml
deanshelton913: Good catch. This limitation was decided against. removed

docs/static/chart/opensource-config.yaml
deanshelton913: fixed

dragonchain/webserver/lib/dragonnet.py
```diff
@@ -80,13 +87,48 @@ def process_receipt_v1(block_dto: Dict[str, Any]) -> None:
...
+                    url,
+                    string_json,
+                    headers={"Content-Type": "application/json", "Authorization": f"Bearer {signature}", "DragonchainId": public_key},
```
cheeseandcereal: I'm confused as this isn't really how bearer tokens work. If you want to pass the signature in the header, I would probably just use a custom header for less confusion

dragonchain/webserver/lib/dragonnet.py
```diff
@@ -80,13 +87,48 @@ def process_receipt_v1(block_dto: Dict[str, Any]) -> None:
...
+                )
+                _log.debug(f"Notify <- {response.status_code} POST {url} {response.text}")
+    except Exception as e:
+        # This feature is fire & forget. All exceptions will just log.
+        _log.error(e)
```
cheeseandcereal: use _log.exception to make sure the traceback is included

dragonchain/webserver/lib/dragonnet.py
cheeseandcereal: Also note the `.export_as_at_rest()` is a potentially changing schema as dragonchain versions change.

If you want to ensure this is a static schema, then you'll have to craft a custom output dictionary, like webserver handlers

CHANGELOG.md
deanshelton913: agreed. Done

docs/static/chart/opensource-config.yaml
deanshelton913: fixed

dragonchain/webserver/lib/dragonnet.py
deanshelton913: updated

dragonchain/webserver/lib/dragonnet.py
deanshelton913: done

dragonchain/webserver/lib/dragonnet.py
deanshelton913: spoke IRL. We are cool with this

deanshelton913:
Update: We are hard at work on the billing feature for managed services. I do plan on coming back to this PR shortly

dragonchain/webserver/lib/dragonnet.py
deanshelton913: (26 days later) Agreed. 
haha. 
After the release of billing, I can jump back into this. Yes. I agree. This PR has been updated to move this logic into the broadcast processor.

CHANGELOG.md
```diff
@@ -48,6 +48,8 @@ for more details on upgrading to v4.
   - Refactor all query endpoints, including changing query inputs (Breaking Change)
   - Changed route for `GET /v1/contract` from a query to a list (Breaking Change)
   - Add support to disable schedule when updating a smart contract
+  - Adds verification-notification callback in the reciept endpoint
+
```
cheeseandcereal: extra line

dragonchain/broadcast_processor/broadcast_functions.py
```diff
@@ -109,6 +111,24 @@ def increment_storage_error_sync(block_id: str, current_level: int) -> None:
     return await redis.z_range_by_score_async(IN_FLIGHT_KEY, 0, int(time.time()), withscores=True, offset=0, count=1000)
 
 
+async def get_notification_verifications_for_broadcast_async() -> List[str]:
+    """Get the notifications scheduled to be checked by the broadcast processor right now
+    Args:
```
cheeseandcereal: kill this line

dragonchain/broadcast_processor/broadcast_functions.py
```diff
@@ -109,6 +111,24 @@ def increment_storage_error_sync(block_id: str, current_level: int) -> None:
...
+    Returns:
+        List of notifications to be re-transmitted by the broadcast processor (limit of 1000)
+    """
+    return await redis.lrange_async(NOTIFICATION_KEY, 0, 1000)
+
+
+async def remove_notification_verification_for_broadcast_async(value: str) -> int:
+    """Removes the notification object
+    Args:
```
cheeseandcereal: should be filled in

dragonchain/broadcast_processor/broadcast_functions.py
```diff
@@ -176,6 +196,27 @@ def schedule_block_for_broadcast_sync(block_id: str, time: int = 0) -> None:
     await redis.zadd_async(IN_FLIGHT_KEY, time, block_id)
 
 
+def schedule_notification_for_broadcast_sync(notification_location: str, time: int = 0) -> None:
```
cheeseandcereal: should remove time from this function as it's now irrelevant.

dragonchain/broadcast_processor/broadcast_functions.py
```diff
@@ -176,6 +196,27 @@ def schedule_block_for_broadcast_sync(block_id: str, time: int = 0) -> None:
...
+    """Schedule a certain notification to be checked by the broadcast processor (sync)
+    Args:
+        notification_location: notification_location to schedule
+        time: unix timestamp to re-schedule for checking (0 means ASAP)
+    """
+    redis.lpush_sync(NOTIFICATION_KEY, notification_location)
+
+
+def verification_storage_location(l1_block_id: str, level_received_from, chain_id) -> str:
```
cheeseandcereal: prefer type annotations in the actual signature rather than the docstring

dragonchain/broadcast_processor/broadcast_functions_utest.py
```diff
@@ -227,10 +227,11 @@ def test_set_record_for_block_sync_raises_when_not_accepting_level(self, mock_ge
             exceptions.NotAcceptingVerifications, broadcast_functions.set_receieved_verification_for_block_from_chain_sync, "block_id", 2, "chain_id"
         )
 
+    @patch("dragonchain.broadcast_processor.broadcast_functions.redis.lpush_sync")
     @patch("dragonchain.broadcast_processor.broadcast_functions.dragonnet_config.DRAGONNET_CONFIG", {"l2": {"nodesRequired": 3}})
     @patch("dragonchain.broadcast_processor.broadcast_functions.verifications_key", return_value="key")
     @patch("dragonchain.broadcast_processor.broadcast_functions.get_current_block_level_sync", return_value=2)
-    def test_set_record_for_block_sync_calls_redis_with_correct_params(self, mock_get_block_level, mock_key):
+    def test_set_record_for_block_sync_calls_redis_with_correct_params(self, mock_get_block_level, mock_key, mock_zadd):
```
cheeseandcereal: these should probably be mock_lpush, not zadd (x3 for this file)

dragonchain/broadcast_processor/broadcast_processor.py
```diff
@@ -110,6 +120,74 @@ def make_broadcast_futures(session: aiohttp.ClientSession, block_id: str, level:
     return broadcasts
 
 
+def get_level_from_storage_location(storage_location: str):
```
cheeseandcereal: should annotate return type here for completeness. Would be `Optional[str]`

dragonchain/broadcast_processor/broadcast_processor.py
```diff
@@ -110,6 +120,74 @@ def make_broadcast_futures(session: aiohttp.ClientSession, block_id: str, level:
...
+    Retrieves verifications that need to be processed and sends http requests
+
+    Args:
+        session: aiohttp session for http requests
+    """
+    if VERIFICATION_NOTIFICATION:
+        futures = set()
+        for storage_location in await broadcast_functions.get_notification_verifications_for_broadcast_async():
+            verification_bytes = storage.get(storage_location)
+            level = get_level_from_storage_location(storage_location)
```
cheeseandcereal: you don't check if this ever comes back none.

This would've been caught by mypy had the return type annotation existed for the `get_level_from_storage_location` method

dragonchain/broadcast_processor/broadcast_processor.py
```diff
@@ -110,6 +120,74 @@ def make_broadcast_futures(session: aiohttp.ClientSession, block_id: str, level:
...
+
+    Returns:
+        None
+    """
+    _log.debug(f"Notification -> {url}")
+    resp = await session.post(
+        url=url, data=verification_bytes, headers={"dragonchainId": keys.get_public_id(), "signature": signature}, timeout=HTTP_REQUEST_TIMEOUT
+    )
+    _log.debug(f"Notification <- {resp.status} {url}")
+    await broadcast_functions.remove_notification_verification_for_broadcast_async(redis_list_value)
```
cheeseandcereal: removing a specific value from a redis list is O(n) where n is the size of the list. If we're doing this operation for literally every single broadcast, maybe we should consider having this be a set if possible, as removing a member from a set in redis is O(1) ?

Just a thought

dragonchain/lib/database/redis.py
```diff
@@ -161,6 +161,16 @@ def _initialize_redis(host: str, port: int, wait_time: int = 30) -> redis.Redis:
     )
 
 
+async def lrange_async(key: str, start: int, end: int) -> List:
```
cheeseandcereal: Probably need a decode flag here

dragonchain/webserver/lib/dragonnet.py
```diff
@@ -69,7 +69,8 @@ def process_receipt_v1(block_dto: Dict[str, Any]) -> None:
         validations = matchmaking.get_claim_check(l1_block_id)["validations"][f"l{level_received_from}"]
         if (block_model.dc_id in validations) and broadcast_functions.is_block_accepting_verifications_from_level(l1_block_id, level_received_from):
             _log.info(f"Verified that block {l1_block_id} was sent. Inserting receipt")
-            storage.put_object_as_json(f"{FOLDER}/{l1_block_id}-l{level_received_from}-{block_model.dc_id}", block_model.export_as_at_rest())
+            storage_location = broadcast_functions.verification_storage_location(l1_block_id, level_received_from, block_model.dc_id)
```
cheeseandcereal: i like this

helm/opensource-config.yaml
```diff
@@ -86,6 +86,34 @@ dragonchain:
...
+  # Authentication: Bearer ${signature}
+  # DragonchainId: PublicIdOfThisL1Dragonchain
+  # Content-Type: application/json
+  # -- Body --
+  # The post body will be JSON of the raw verification object received by this dragonchain. See Dragonchain documentation for details.
+  # -- Signature --
+  # The bearer token will be the signature of the utf8 encoded bytes of the post body signed with the private key, and verifiable with
+  # the public key found in the DragonchainId header. NOTE: Do not trust this header alone to verify signature. Make sure it's a public id
+  # that you trust and expect.
+  verificationNotification: null
```
cheeseandcereal: where's the counterpart for this to actually insert this into the env vars of the chain correctly? I don't see any edited helm file

cheeseandcereal: Notes: you also will need a default in the `values.yaml` of the chart itself for strict helm linting after you edit the helm chart.

dragonchain/broadcast_processor/broadcast_functions.py
```diff
@@ -109,6 +111,24 @@ def increment_storage_error_sync(block_id: str, current_level: int) -> None:
...
+    return await redis.lrange_async(NOTIFICATION_KEY, 0, 1000)
+
+
+async def remove_notification_verification_for_broadcast_async(value: str) -> int:
+    """Removes the notification object
+    Args:
+    Returns:
+        Number of elements removed from the notification verification list
+    """
+    return await redis.lrem_async(NOTIFICATION_KEY, value)
```
cheeseandcereal: I think this is wrong. If your goal is to remove a single instance of `value` in the redis list, you definitely have to pass in a count of `1` somewhere on top of your value.

dragonchain/broadcast_processor/broadcast_processor.py
deanshelton913: good idea. changing

dragonchain/broadcast_processor/broadcast_processor.py
deanshelton913: true that. also typed that function. change coming up

dragonchain/broadcast_processor/broadcast_processor.py
deanshelton913: Good call. Actually swapping this to a set should be pretty simple

dragonchain/lib/database/redis.py
deanshelton913: imma kill this and replace with set stuff

helm/opensource-config.yaml
deanshelton913: yup. Lost in a rebase. I have them coming up

dragonchain/broadcast_processor/broadcast_functions_utest.py
deanshelton913: even better -> mock_sadd

CHANGELOG.md
deanshelton913: fixed

dragonchain/broadcast_processor/broadcast_functions.py
deanshelton913: killed

dragonchain/broadcast_processor/broadcast_functions.py
deanshelton913: done

dragonchain/broadcast_processor/broadcast_functions.py
deanshelton913: done

dragonchain/broadcast_processor/broadcast_functions.py
deanshelton913: agree

dragonchain/broadcast_processor/broadcast_processor.py
deanshelton913: fixed in the `get_all_notification_endpoints` above

dragonchain/broadcast_processor/broadcast_processor.py
deanshelton913: swapped

dragonchain/broadcast_processor/broadcast_functions.py
```diff
@@ -102,13 +105,30 @@ def increment_storage_error_sync(block_id: str, current_level: int) -> None:
...
     Returns:
         List of (block_id, score) tuples to be processed by the broadcast processor (limit of 1000)
     """
     return await redis.z_range_by_score_async(IN_FLIGHT_KEY, 0, int(time.time()), withscores=True, offset=0, count=1000)
 
 
+async def get_notification_verifications_for_broadcast_async() -> set:
+    """Get the notifications scheduled to be checked by the broadcast processor right now
+    Returns:
+        List of notifications to be re-transmitted by the broadcast processor (limit of 1000)
```
cheeseandcereal: incorrect limit. Also this method probably needs a decode since it's getting data out of redis. Also I'm not 100% sure this returns a set. I think the other redis library returns a list for smembers, and we have to coerce it. I would double check and confirm.

dragonchain/broadcast_processor/broadcast_functions.py
```diff
@@ -197,6 +237,10 @@ def set_receieved_verification_for_block_from_chain_sync(block_id: str, level: i
     verifications = p.execute()[1]  # Execute the commands and get the result of the scard operation (number of members in the set)
     required = dragonnet_config.DRAGONNET_CONFIG[f"l{level}"]["nodesRequired"]
 
+    if os.environ.get("VERIFICATION_NOTIFICATION") is not None:
```
cheeseandcereal: ideally we abstract this to a module level variable so we only have to retrieve the env var once

dragonchain/broadcast_processor/broadcast_processor.py
```diff
@@ -110,6 +120,79 @@ def make_broadcast_futures(session: aiohttp.ClientSession, block_id: str, level:
...
+        futures = set()
+        for storage_location in await broadcast_functions.get_notification_verifications_for_broadcast_async():
+            verification_bytes = storage.get(storage_location)
+            level = get_level_from_storage_location(storage_location)
+            signature = sign(verification_bytes)
+            for url in get_all_notification_endpoints(level):
+                future = send_notification_verification(session, url, verification_bytes, signature, storage_location)
+                futures.add(asyncio.ensure_future(future))
+        try:
+            await asyncio.gather(*futures)  # "Fire & forget"
```
cheeseandcereal: if you get one error, it will not send the rest of the verifications. Not sure if this is intentional.

If you want it to try for all futures regardless of which ones error, add `return_exceptions=True` to the asyncio.gather call

dragonchain/broadcast_processor/broadcast_functions.py
deanshelton913: checked. Confirmed. Is a set

dragonchain/broadcast_processor/broadcast_functions.py
deanshelton913: agreed. done

dragonchain/broadcast_processor/broadcast_processor.py
deanshelton913: good call. Updated

dragonchain/broadcast_processor/broadcast_processor.py
```diff
@@ -110,6 +120,81 @@ def make_broadcast_futures(session: aiohttp.ClientSession, block_id: str, level:
...
+            verification_bytes = storage.get(storage_location)
+            level = get_level_from_storage_location(storage_location)
+            if level is None:
+                _log.error(f"Unable to parse level value from string {storage_location}. Removing verification notification from set.")
+                broadcast_functions.remove_notification_verification_for_broadcast_async(storage_location)
+                continue
+            signature = sign(verification_bytes)
+            for url in get_all_notification_endpoints(level):
+                future = send_notification_verification(session, url, verification_bytes, signature, storage_location)
+                futures.add(asyncio.ensure_future(future))
```
cheeseandcereal: prefer `asyncio.create_task`

https://docs.python.org/3/library/asyncio-task.html#creating-tasks

dragonchain/broadcast_processor/broadcast_processor.py
```diff
@@ -110,6 +120,81 @@ def make_broadcast_futures(session: aiohttp.ClientSession, block_id: str, level:
...
+        try:
+            await asyncio.gather(*futures, return_exceptions=True)  # "Fire & forget"
+        except Exception:
```
cheeseandcereal: you don't need this try/except if you have return_exceptions=True

asyncio.gather should never throw an exception in this configuration unless something is seriously wrong

dragonchain/lib/database/redis.py
```diff
@@ -161,6 +161,16 @@ def _initialize_redis(host: str, port: int, wait_time: int = 30) -> redis.Redis:
     )
 
 
+async def lrem_async(key: str, value: str, count=1) -> int:
```
cheeseandcereal: i don't think this function is used anymore

cheeseandcereal:
Please add the boilerplate tests for redis.py for the couple functions that you added.

Other than that and my few remaining comments, looks good.

dragonchain/lib/database/redis.py
```diff
@@ -161,6 +161,11 @@ def _initialize_redis(host: str, port: int, wait_time: int = 30) -> redis.Redis:
     )
 
 
+async def srem_async(key: str, value: str) -> int:
```
cheeseandcereal: add test for this

dragonchain/broadcast_processor/broadcast_functions.py
```diff
@@ -102,13 +106,30 @@ def increment_storage_error_sync(block_id: str, current_level: int) -> None:
...
 async def get_blocks_to_process_for_broadcast_async() -> List[Tuple[str, int]]:
     """Get the blocks scheduled to be checked by the broadcast processor right now
-    Args:
     Returns:
         List of (block_id, score) tuples to be processed by the broadcast processor (limit of 1000)
     """
     return await redis.z_range_by_score_async(IN_FLIGHT_KEY, 0, int(time.time()), withscores=True, offset=0, count=1000)
 
 
+async def get_notification_verifications_for_broadcast_async() -> set:
```
cheeseandcereal: add tests for all the new functions in this file

cheeseandcereal:
Add documentation about the broadcast processor details for its new functionality

dragonchain/broadcast_processor/broadcast_functions.py
deanshelton913: done

dragonchain/lib/database/redis.py
deanshelton913: done

PR #217, opened 2019-10-14
Add transaction invoker indexing

cheeseandcereal:
## Description

Adds a redisearch tag field "invoker" on new transaction types/smart contracts which is populated when indexing a transaction that has an invoker field (smart contract response transactions).

Although I've done some preliminary testing, this requires a bit more manual testing to ensure this isn't a breaking change with existing chains/indexes.

##### Checklist


cheeseandcereal:
Manually testing now

cheeseandcereal:
Everything appears to work with this PR.

Existing smart contracts won't have the invoker tag in their index, but they will be accepted in the index document regardless.

Previous transactions can be re-indexed for this field with the added docs.

New smart contracts work fine.

PR #235, opened 2019-11-07
Bnb fixes

Wingman4l7:
##### Checklist

- [x] tests are included


dragonchain/lib/dto/bnb.py
```diff
@@ -275,36 +275,47 @@ def _fetch_account(self):
...
+            symbol (str): the exchange symbol for the token
+            memo (str): string of data to publish in the transaction
+        Returns:
+            dict of the constructed transaction
+        """
+        amount = raw_transaction["amount"]
+        if amount <= 0:
+            raise exceptions.BadRequest("[BINANCE] Amount in transaction cannot be less than or equal to 0.")
 
+        symbol = raw_transaction["symbol"]
```
cheeseandcereal: you assume symbol exists, but it's not required in your schema

Wingman4l7: It's not required, but it is defaulted to "BNB" if it is not provided -- see `dragonchain_client.py`.

cheeseandcereal: yeah, but the sdk isn't how you should rely on a default. What if someone else writes a dragonchain client? Now the Dragonchain can return a 500 error if you don't know that you have to provide a symbol, which isn't required by the schema on Dragonchain.

You would expect not providing some part of the schema to result in a 400, not an unhandled 500.

Defaulting a value in the SDK is a nice-to-have, but is not a substitute for making sure the back-end accepts and handles good/bad/missing values.

An api is a user interface too. You definitely shouldn't make the assumption that everyone will interact with the Dragonchain via only our sanctioned SDKs, and the backend should be able to handle bad input.

So we need to either require symbol in the schema, or default it on the back-end if it's not provided.

Wingman4l7: Made sure that `_build_transaction_msg()` function in the backend provided the proper defaults.  Removed defaults from SDK, feels redundant there.

dragonchain/lib/dto/bnb.py
```diff
@@ -275,36 +275,46 @@ def _fetch_account(self):
...
+            symbol (str, optional): the exchange symbol for the token (defaults to 'BNB')
+            memo (str, optional): string of data to publish in the transaction (defaults to '')
+        Returns:
+            dict of the constructed transaction
+        """
+        amount = raw_transaction["amount"]
+        if amount <= 0:
+            raise exceptions.BadRequest("[BINANCE] Amount in transaction cannot be less than or equal to 0.")
+
+        symbol = raw_transaction["symbol"] if raw_transaction.get("symbol") is not None else "BNB"
```
cheeseandcereal: shorthand:
```suggestion
        symbol = raw_transaction.get("symbol") or "BNB"
```

dragonchain/lib/dto/bnb.py
```diff
@@ -275,36 +275,46 @@ def _fetch_account(self):
...
+        memo = raw_transaction["memo"] if raw_transaction.get("memo") is not None else ""
```
cheeseandcereal: shorthand:
```suggestion
        memo = raw_transaction.get("memo") or ""
```

dragonchain/lib/dto/bnb.py
Wingman4l7: Fixed!

dragonchain/lib/dto/bnb.py
Wingman4l7: Fixed!

PR #237, opened 2019-11-09
Fix opensource l1 job processor

### LINKED ISSUE #236: L1 Job Processor Crashes when IAM_ROLE environment variable is not defined
- **Dragonchain Version**: 4.1.0
- **Entrypoint**: job_processor

When using the job processor on an L1 chain, it will simply crash if you haven't defined the IAM_ROLE environment variable, which shouldn't be required.

cheeseandcereal:
Fixes #236

## Description

Fixes issue with L1 job processor crashing due to bad environment variable

Also fixes small error with unused value in helm chart

##### Checklist

- [x] tests are included


dragonchain/job_processor/job_processor.py
```diff
@@ -32,7 +32,7 @@
 INTERNAL_ID = os.environ["INTERNAL_ID"]
 STAGE = os.environ["STAGE"]
 REGISTRY = os.environ["REGISTRY"]
-IAM_ROLE = os.environ["IAM_ROLE"]
+IAM_ROLE = os.environ.get("IAM_ROLE") or "N/A"
```
regan-karlewicz: I'm all for the `.get`, but not sure why we would hardcode `N/A` there. Why not just have no annotations if the environment variable doesn't exist?

cheeseandcereal: agreed. fixed

PR #255, opened 2019-11-20
Fixes issue where claims aren't being sent to matchmaking.

Wingman4l7:
## Description

Fixes issue where L1 claims are not sent to matchmaking if matchmaking is down during the sending of receipts to lower level nodes.
See:  https://dragondev.atlassian.net/browse/DRGN-282

##### Checklist

- [x] tests are included

### DIFF
--- a/dragonchain/exceptions.py
+++ b/dragonchain/exceptions.py
@@ -167,5 +167,9 @@ class MatchmakingError(DragonchainException):
     """Exception raised by matchmaking client when a problem has occurred"""
 
 
+class MatchmakingRetryableError(DragonchainException):
+    """Exception raised by matchmaking when there is a server error"""
+
+
 class PartyError(DragonchainException):
     """Exception raised by party client when a problem has occurred"""
--- a/dragonchain/lib/broadcast.py
+++ b/dragonchain/lib/broadcast.py
@@ -22,8 +22,10 @@ from typing import TYPE_CHECKING, cast
 import requests
 
 from dragonchain import logger
+from dragonchain import exceptions
 from dragonchain.lib import authorization
 from dragonchain.lib import matchmaking
+from dragonchain.lib.database import redis
 
 if TYPE_CHECKING:
     from dragonchain.lib.dto import l5_block_model
@@ -99,8 +101,11 @@ def send_receipts(l5_block: "l5_block_model.L5BlockModel") -> None:
                 try:
                     claim_check_id = f"{chain_id}-{block}"
                     matchmaking.resolve_claim_check(claim_check_id)
+                except exceptions.MatchmakingRetryableError:  # any 500-level server errors
+                    _log.exception(f"Adding claim to failed queue.  Claim ID: {claim_check_id}")
+                    redis.sadd_sync("mq:failed-claims", claim_check_id)  # using a set avoids duplicates
                 except Exception:
-                    _log.exception("Failure to finalize claim in matchmaking. Sending reciepts to lower level nodes.")
+                    _log.exception("Failure to finalize claim in matchmaking. Sending receipts to lower level nodes.")
         except Exception as e:
             _log.exception(f"[BROADCAST] Error while trying to broadcast down for l4 block {l4_block}\n{e}\n!Will ignore this broadcast!")
 
--- a/dragonchain/lib/matchmaking.py
+++ b/dragonchain/lib/matchmaking.py
@@ -337,6 +337,8 @@ def make_matchmaking_request(
             raise exceptions.InsufficientFunds("received insufficient funds (402) from matchmaking")
         elif response.status_code == 404:
             raise exceptions.NotFound("Not found (404) from matchmaking")
+        elif response.status_code >= 500:
+            raise exceptions.MatchmakingRetryableError(f"[MATCHMAKING] Server error {response.status_code} from matchmaking")
         raise exceptions.MatchmakingError(
             f"Received unexpected response code {response.status_code} from matchmaking with response:\n{response.text}"
         )
--- a/dragonchain/transaction_processor/level_5_actions.py
+++ b/dragonchain/transaction_processor/level_5_actions.py
@@ -37,6 +37,9 @@ from dragonchain.lib.dao import interchain_dao
 from dragonchain.lib import queue
 from dragonchain.transaction_processor import shared_functions
 from dragonchain.lib.database import redisearch
+from dragonchain.lib.database import redis
+
+FAILED_CLAIMS_KEY = "mq:failed-claims"
 
 if TYPE_CHECKING:
     from dragonchain.lib.dto import model  # noqa: F401
@@ -78,6 +81,7 @@ def execute() -> None:
     * Publishes to public nodes when required
     * Locates confirmations from public nodes when required
     * Sends receipts to all L1 blocks represented in public broadcast
+    * Finalizes any previously failed claims present in backlog
     """
     matchmaking.renew_registration_if_necessary()
     # Check if there are any funds
@@ -106,6 +110,22 @@ def execute() -> None:
 
     # Only check confirmations if there are any pending
     check_confirmations()
+
+    # go through the failed claims backlog
+    _log.info("Trying to finalize any previously failed claims...")
+    process_claims_backlog()
+
+
+def process_claims_backlog() -> None:
+    claims_set = redis.smembers_sync(FAILED_CLAIMS_KEY)  # used sadd to make a set
+    _log.info(f"Claim backlog count: {len(claims_set)}")
+    for claim_check_id in claims_set:
+        try:
+            matchmaking.resolve_claim_check(claim_check_id)
+        except Exception:
+            _log.exception("Failure to finalize claim in matchmaking.  Skipping the rest of the retry queue.")
+            return  # short-circuit and exit early, matchmaking still unreachable
+        redis.srem_sync(FAILED_CLAIMS_KEY, claim_check_id)  # claim successfully finalized, remove from backlog
 
 
 def broadcast_clean_up(l5_block: l5_block_model.L5BlockModel) -> None:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 4.2.1
 
+- **Bugs:**
+  - Fix a bug where a failure in matchmaking would result in claims not being finalized
 - **Documentation:**
   - Update documentation links to add the [getting started guide](https://docs.dragonchain.com/)
 - **Packaging:**
@@ -50,7 +52,7 @@ field where necessary.
 - **Feature:**
   - Default error reporting to save to disk, so that crash logs/tracebacks can be automatically saved
   - Provide better error message when bad input to api doesn't match required schemas
-  - Adds verification-notification callback in the reciept endpoint
+  - Adds verification-notification callback in the receipt endpoint
   - Add indexed redisearch tag field "invoker" by default when indexing smart contract transactions
   - Remove max limit for custom indexes on a transaction type/smart contract
 - **Bugs:**
--- a/dragonchain/transaction_processor/level_1_actions.py
+++ b/dragonchain/transaction_processor/level_1_actions.py
@@ -46,7 +46,7 @@ def execute() -> None:
     if BROADCAST:
         try:
             matchmaking.renew_registration_if_necessary()
-        except exceptions.MatchmakingError:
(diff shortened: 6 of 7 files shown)


dragonchain/lib/broadcast.py
```diff
@@ -36,6 +37,8 @@
 
 def dispatch(block: "model.BlockModel") -> None:
     if LEVEL == "5":
+        _log.info("Trying to finalize any previously failed claims...")
+        process_claims_backlog()
```
cheeseandcereal: You can't just make an async call. All this does is return a coroutine. This will do literally nothing.

I think you need to look up how to use asyncio and python...

L5_deployment_testing.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/sh
```
cheeseandcereal: don't commit this file

.version
```diff
@@ -1 +1 @@
-4.2.0
\ No newline at end of file
+gould-4.2.0
```
cheeseandcereal: don't commit this version

dragonchain/lib/broadcast.py
```diff
@@ -60,6 +63,22 @@ def send_receipt(block: "model.BlockModel") -> None:
...
+        claim_list.append(value)
+    _log.info(f"Claim backlog count: {len(claim_list)}")
+    if claim_list:
+        for claim_check_id in claim_list:
+            try:
+                matchmaking.resolve_claim_check(claim_check_id)
+            except Exception:
+                _log.exception("Failure to finalize claim in matchmaking.  Skipping the rest of the retry queue.")
+                return  # short-circuit and exit early, matchmaking still unreachable
+            redis.hdel_async("mq:failed-claims", claim_check_id)  # claim successfully finalized, remove from backlog
```
cheeseandcereal: why are you calling an async function without await? This will not work

dragonchain/lib/broadcast.py
```diff
@@ -100,7 +119,9 @@ def send_receipts(l5_block: "l5_block_model.L5BlockModel") -> None:
                     claim_check_id = f"{chain_id}-{block}"
                     matchmaking.resolve_claim_check(claim_check_id)
                 except Exception:
-                    _log.exception("Failure to finalize claim in matchmaking. Sending reciepts to lower level nodes.")
+                    _log.exception("Failure to finalize claim in matchmaking. Sending receipts to lower level nodes.")
+                    _log.exception(f"Adding claim to failed queue.  Claim ID: {claim_check_id}")
```
cheeseandcereal: don't make 2 calls to the logger. Not only will this print a stack trace twice, it's unnecessary overhead calling the logger again when all you want is a newline

cheeseandcereal:
You need to add a changelog entry

L5_deployment_testing.sh
Wingman4l7: yes, had to leave this in for pending test for binance, will be removed in final commit

.version
Wingman4l7: yes, had to leave this in for pending test for binance, will be removed in final commit

dragonchain/lib/broadcast.py
Wingman4l7: Good point.  Fixed.

Wingman4l7:
> You need to add a changelog entry

Added.

dragonchain/lib/broadcast.py
Wingman4l7: Didn't mean to be mixing these two, auto-complete fail.

dragonchain/lib/broadcast.py
Wingman4l7: Don't know why I originally set this up with some async stuff, I don't see a particular need for it.  Wasn't ready for PR review, fixed.

dragonchain/lib/broadcast.py
```diff
@@ -100,7 +118,11 @@ def send_receipts(l5_block: "l5_block_model.L5BlockModel") -> None:
                     claim_check_id = f"{chain_id}-{block}"
                     matchmaking.resolve_claim_check(claim_check_id)
                 except Exception:
-                    _log.exception("Failure to finalize claim in matchmaking. Sending reciepts to lower level nodes.")
+                    _log.exception(
+                        "Failure to finalize claim in matchmaking. Sending receipts to lower level nodes. \n"
+                        f"Adding claim to failed queue.  Claim ID: {claim_check_id}"
+                    )
+                    redis.lpush_sync("mq:failed-claims", claim_check_id)
```
cheeseandcereal: this needs to be a confirmed retry-able exception. We can't just blindly retry.

For example, if we got back an HTTP 400 from matchmaking, we should NOT retry, and should rather raise the exception since there is NO reason an L5 should ever get a 400 from matchmaking, and if it did, something is wrong, and we definitely shouldn't retry.

dragonchain/lib/broadcast.py
cheeseandcereal: it doesn't make sense to even have this here. This will only process a backlog every time an L5 actually transmits a block. That means it might only retry once every 2 or 3 hours. This should be retried every time the transaction processor runs, not just when it happens to be broadcasting a block

dragonchain/lib/broadcast.py
Wingman4l7: Created exception type for 500-level errors and restricted queue additions to these exceptions only.

dragonchain/lib/broadcast.py
Wingman4l7: Moved function to `level_5_actions`, makes more sense for it to belong there and be invoked there.

.version
Wingman4l7: Binance testing done, fixed.

L5_deployment_testing.sh
Wingman4l7: Binance testing done, fixed.

cheeseandcereal:
Your diff looks weird because it needs a rebase (not just merging master into this branch)

dragonchain/exceptions.py
```diff
@@ -167,5 +167,9 @@ class MatchmakingError(DragonchainException):
     """Exception raised by matchmaking client when a problem has occurred"""
 
 
+class RetryableError(DragonchainException):
```
cheeseandcereal: If this is specifically for matchmaking, it should probably be something like `MatchmakingRetryableError`

Also anywhere that used to do `except MatchmakingError` will now need to also catch this error too

dragonchain/transaction_processor/level_5_actions.py
```diff
@@ -107,6 +109,25 @@ def execute() -> None:
     # Only check confirmations if there are any pending
     check_confirmations()
 
+    # go through the failed claims backlog
+    _log.info("Trying to finalize any previously failed claims...")
+    process_claims_backlog()
+
+
+def process_claims_backlog() -> None:
+    claims = redis.hgetall_sync("mq:failed-claims", decode=False)
```
cheeseandcereal: hgetall? that's definitely not right, also why no decode?

dragonchain/transaction_processor/level_5_actions.py
```diff
@@ -107,6 +109,25 @@ def execute() -> None:
...
+    for _, value in claims.items():
+        claim_list.append(value)
+    _log.info(f"Claim backlog count: {len(claim_list)}")
+    for claim_check_id in claim_list:
+        try:
+            matchmaking.resolve_claim_check(claim_check_id)
+        except Exception:
+            _log.exception("Failure to finalize claim in matchmaking.  Skipping the rest of the retry queue.")
+            return  # short-circuit and exit early, matchmaking still unreachable
+        redis.hdel_sync("mq:failed-claims", claim_check_id)  # claim successfully finalized, remove from backlog
```
cheeseandcereal: hdel? that's definitely not right for a redis list.

dragonchain/exceptions.py
Wingman4l7: Good point, I'll rename it.  
Looks like the only other spots catching `MatchmakingError` exceptions are `execute()` in `level_1_actions.py` and `update_registration()` in `matchmaking.py` ...

dragonchain/transaction_processor/level_5_actions.py
```diff
@@ -107,6 +109,25 @@ def execute() -> None:
     # Only check confirmations if there are any pending
     check_confirmations()
 
+    # go through the failed claims backlog
+    _log.info("Trying to finalize any previously failed claims...")
+    process_claims_backlog()
+
+
+def process_claims_backlog() -> None:
```
cheeseandcereal: this function definitely needs tests. Just looking at it, I can tell it won't run

dragonchain/transaction_processor/level_5_actions.py
Wingman4l7: Should be using set.  Fixed.

dragonchain/transaction_processor/level_5_actions.py
Wingman4l7: Should be using set.  Fixed.

dragonchain/transaction_processor/level_5_actions.py
Wingman4l7: Adding tests.

dragonchain/lib/matchmaking.py
```diff
@@ -329,14 +329,18 @@ def make_matchmaking_request(
             return make_matchmaking_request(http_verb=http_verb, path=path, json_content=json_content, retry=False, authenticated=authenticated)
         elif retry and response.status_code == 403 and authenticated:
             _log.warning(
-                "[MATCHMAKING] received 403 from matchmaking. Registration is expired or dragon net config is invalid. Re-registering and trying again"
+                "[MATCHMAKING] received 403 from matchmaking. "
```
cheeseandcereal: not sure why you changed this log

Wingman4l7: It was passing the linter (I guess we ignore PEP rules on line length?), but I'm biased against output strings I can't read without having a 30" wide monitor >_>  I can change it back if you want.

dragonchain/lib/matchmaking.py
```diff
@@ -329,14 +329,18 @@ def make_matchmaking_request(
...
+                "Re-registering and trying again"
             )
             register(retry=False)
             return make_matchmaking_request(http_verb=http_verb, path=path, json_content=json_content, retry=False, authenticated=authenticated)
         elif response.status_code == 402:
             raise exceptions.InsufficientFunds("received insufficient funds (402) from matchmaking")
         elif response.status_code == 404:
             raise exceptions.NotFound("Not found (404) from matchmaking")
+        elif response.status_code >= 500:
+            raise exceptions.MatchmakingRetryableError(f"[MATCHMAKING] Server error {response.status_code} from matchmaking")
```
cheeseandcereal: Anywhere where anything was catching an `exceptions.MatchmakingError` probably needs to be able to catch this new new code too

dragonchain/transaction_processor/level_5_actions.py
```diff
@@ -107,6 +109,22 @@ def execute() -> None:
...
+    claims_set = redis.smembers_sync("mq:failed-claims")  # used sadd to make a set
```
cheeseandcereal: shouldn't be hard-coding strings like "mq:failed-claims". Make this as a constant in the module, or something

dragonchain/lib/matchmaking.py
Wingman4l7: yeah, there was one spot in... L1 actions I think, which was catching a `MatchmakingError`, I added `MatchmakingRetryableError` to it

dragonchain/lib/matchmaking.py
cheeseandcereal: The formatter enforces line length. It's also hard to tell that these lines are necessarily connected when reading line-seperated logs.

dragonchain/lib/matchmaking.py
cheeseandcereal: Never mind, you got it

dragonchain/lib/matchmaking.py
Wingman4l7: That formatting doesn't actually break up the lines in the log, only in the source code.  I'll just revert it, not a hill I care about dying on. :P

dragonchain/transaction_processor/level_5_actions.py
Wingman4l7: Good point.  Fixed.

cheeseandcereal:
Looks good, lets just see this working on a real chain before merging

PR #256, opened 2019-11-20
Release 4.2.0

cheeseandcereal:
## Description

Dragonchain release 4.2.0 with binance interchain support and various bugfixes for on-prem l1 deployments.

Proper tagged release will be created after the merge.


dragonchain/lib/segwit_addr.py
```diff
@@ -0,0 +1,126 @@
+# Copyright (c) 2017 Pieter Wuille
```
rdaquilante: What's this?

cheeseandcereal: A vendored dependency

rdaquilante: Oh this was the file we needed, but we didn't want the whole dependency. I remember now.

PR #268, opened 2019-12-06
implementing deadline for L5s

Wingman4l7:
##### Checklist

- [x] tests are included
- [X] changelog has been modified for the changes

### DIFF
--- a/dragonchain/broadcast_processor/broadcast_processor.py
+++ b/dragonchain/broadcast_processor/broadcast_processor.py
@@ -110,7 +110,7 @@ def set_l5_wait_time(chain_id: str) -> int:
     except Exception:  # if there is an error when contacting matchmaking
         _log.exception(f"[BROADCAST PROCESSOR] Exception when fetching config from matchmaking for chain {chain_id}")
         broadcast_receipt_wait_time_l5 = 43200  # seconds (12 hours) [fallback value]
-    L5_WAIT_TIMES[chain_id] = broadcast_receipt_wait_time_l5  # adds to module-level dictionary cache
+    _l5_wait_times[chain_id] = broadcast_receipt_wait_time_l5  # adds to module-level dictionary cache
     return broadcast_receipt_wait_time_l5
 
 
--- a/dragonchain/broadcast_processor/broadcast_processor.py
+++ b/dragonchain/broadcast_processor/broadcast_processor.py
@@ -35,19 +35,18 @@ from dragonchain.lib import crypto
 from dragonchain import logger
 from dragonchain import exceptions
 from dragonchain.lib.interfaces import storage
+from dragonchain.lib.dto import eth, btc, bnb
 
 BROADCAST = os.environ["BROADCAST"]
 LEVEL = os.environ["LEVEL"]
 HTTP_REQUEST_TIMEOUT = 30  # seconds
 BROADCAST_RECEIPT_WAIT_TIME = 35  # seconds
-# TODO Make L5 wait time dynamic
-BROADCAST_RECEIPT_WAIT_TIME_L5 = 43200  # seconds
 
 VERIFICATION_NOTIFICATION: Dict[str, List[str]] = {}
 if os.environ.get("VERIFICATION_NOTIFICATION") is not None:
     VERIFICATION_NOTIFICATION = cast(Dict[str, List[str]], json.loads(os.environ["VERIFICATION_NOTIFICATION"]))
 
-
+_l5_wait_times: Dict[str, int] = {}  # dcID: wait in seconds
 _log = logger.get_logger()
 # For these variables, we are sure to call setup() when initializing this module before using it, so we ignore type error for None
 _requirements: dict = {}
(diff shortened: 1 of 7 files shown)


dragonchain/broadcast_processor/broadcast_processor.py
```diff
@@ -76,6 +75,24 @@ def needed_verifications(level: int) -> int:
...
+    network_attr = {
+        "btc": {"confirmations": btc.CONFIRMATIONS_CONSIDERED_FINAL, "block_time": btc.AVERAGE_BLOCK_TIME, "delay_buffer": 1.5},
+        "eth": {"confirmations": eth.CONFIRMATIONS_CONSIDERED_FINAL, "block_time": eth.AVERAGE_BLOCK_TIME, "delay_buffer": 1.5},
+        "bnb": {"confirmations": bnb.CONFIRMATIONS_CONSIDERED_FINAL, "block_time": bnb.AVERAGE_BLOCK_TIME, "delay_buffer": 1.5},
+    }  # block_time is in seconds; delay_buffer multiplier is to account for network delays
+    try:
+        mm_config = matchmaking.get_registration(chain_id)
+        interchain_network = mm_config["network"]  # returns: "btc" / "eth" / "bnb"
+        attr = network_attr[interchain_network]
+        broadcast_receipt_wait_time_l5 = int(attr["confirmations"] * attr["block_time"] * attr["delay_buffer"])  # in seconds
```
rdaquilante: We still need to take into account remaining_time_to_broadcast for the nodes. This duration won't be sufficient for the 3 hour wait of a btc L5 + confirmation time.

dragonchain/broadcast_processor/broadcast_processor.py
```diff
@@ -76,6 +75,24 @@ def needed_verifications(level: int) -> int:
     return _requirements[f"l{level}"]["nodesRequired"]
 
 
+def get_l5_wait_time(chain_id: str) -> int:
```
cheeseandcereal: should write some unit tests for this

dragonchain/broadcast_processor/broadcast_processor.py
```diff
@@ -232,6 +251,7 @@ def get_all_notification_endpoints(level: str) -> set:
             await broadcast_functions.save_unfinished_claim(block_id)
             continue
         claim_chains = chain_id_set_from_matchmaking_claim(claim, current_level)
+        broadcast_receipt_wait_time_l5 = get_l5_wait_time(claim["properties"]["metadata"]["properties"]["dcId"])  # see: matchmaking schema
```
cheeseandcereal: definitely shouldn't get this value unless we already know this is an L5

dragonchain/webserver/lib/transactions.py
```diff
@@ -133,9 +133,8 @@ def submit_transaction_v1(transaction: Dict[str, Any], callback_url: Optional[st
 
 def submit_bulk_transaction_v1(bulk_transaction: Sequence[Dict[str, Any]], api_key: "api_key_model.APIKeyModel") -> Dict[str, List[Any]]:
     """
-    Formats, validates and enqueues the transactions in
-    the payload of bulk_transaction
-    Returns dictionary of 2 lists, key "201" are successfully created txn ids and key "400" are transactions that failed to post (entire dto passed in from user)
+    Formats, validates and enqueues the transactions in the payload of bulk_transaction
+    Returns tuple of 2 lists, index 0 is successful txn ids and index 1 are transactions that failed to post (entire dto passed in from user)
```
cheeseandcereal: This returns a dictionary, not a tuple. Also seems unrelated...

Wingman4l7: Not my change, must be from messed up rebase -- can do that again.

dragonchain/broadcast_processor/broadcast_processor.py
Wingman4l7: Good point; added conditional.

dragonchain/broadcast_processor/broadcast_processor.py
Wingman4l7: Done.

dragonchain/broadcast_processor/broadcast_processor.py
Wingman4l7: Fixed, included broadcast interval.

dragonchain/webserver/lib/transactions.py
Wingman4l7: This was somehow included from a rebase I apparently messed up.

dragonchain/broadcast_processor/broadcast_processor.py
```diff
@@ -87,6 +86,27 @@ def chain_id_set_from_matchmaking_claim(claim: dict, level: int) -> Set[str]:
...
+    network_attr = {
+        "btc": {"confirmations": btc.CONFIRMATIONS_CONSIDERED_FINAL, "block_time": btc.AVERAGE_BLOCK_TIME, "delay_buffer": 5.0},
+        "eth": {"confirmations": eth.CONFIRMATIONS_CONSIDERED_FINAL, "block_time": eth.AVERAGE_BLOCK_TIME, "delay_buffer": 1.5},
+        "bnb": {"confirmations": bnb.CONFIRMATIONS_CONSIDERED_FINAL, "block_time": bnb.AVERAGE_BLOCK_TIME, "delay_buffer": 1.5},
+    }  # block_time is in seconds; delay_buffer multiplier is to account for network delays
+    try:
+        mm_config = matchmaking.get_registration(chain_id)
+        interchain_network = mm_config["network"]  # returns: "btc" / "eth" / "bnb"
+        broadcast_interval = mm_config["broadcastInterval"]  # returns: decimal value in hours
+        broadcast_interval = int(((broadcast_interval * 60) * 60))  # converts to int value in seconds
```
cheeseandcereal: why all the parens? Makes it a bit harder to read, especially considering order of operations doesn't matter for multiplication

Wingman4l7: Fair point, was a stylistic choice but it makes more sense just to `* 3600` so I'll change it to that.

dragonchain/broadcast_processor/broadcast_processor.py
```diff
@@ -112,6 +132,8 @@ def make_broadcast_futures(session: aiohttp.ClientSession, block_id: str, level:
             headers, data = authorization.generate_authenticated_request("POST", chain, path, broadcast_dto)
             if level != 5:
                 headers["deadline"] = str(BROADCAST_RECEIPT_WAIT_TIME)
+            else:
+                headers["deadline"] = str(get_l5_wait_time(chain))
```
cheeseandcereal: this means that for every l5 broadcast, you are now retrieving from matchmaking and calculating the l5 wait time twice; once in this function, and once in the main broadcast loop for re-scheduling its check. You should only have this in one spot, especially considering it's making an external call to calculate

Wingman4l7: Fixed, as discussed.

dragonchain/broadcast_processor/broadcast_processor.py
```diff
@@ -35,13 +35,13 @@
...
 from dragonchain.lib.interfaces import storage
+from dragonchain.lib.dto import eth, btc, bnb
 
 BROADCAST = os.environ["BROADCAST"]
 LEVEL = os.environ["LEVEL"]
 HTTP_REQUEST_TIMEOUT = 30  # seconds
 BROADCAST_RECEIPT_WAIT_TIME = 35  # seconds
-# TODO Make L5 wait time dynamic
-BROADCAST_RECEIPT_WAIT_TIME_L5 = 43200  # seconds
+L5_WAIT_TIMES: dict = {}  # initialize
```
cheeseandcereal: Even though this is module-level, we want to indicate this is for internal use only, so prepend the var with an underscore and lowercase it.
```suggestion
_l5_wait_times: Dict[str, int] = {}
```

Wingman4l7: Fixed.

PR #269, opened 2019-12-06
Make redisearch optional on verification nodes

cheeseandcereal:
## Description

This PR makes redisearch an optional component for L2+ nodes.

If redisearch is disabled, then all query endpoints are also disabled. Re-enabling redisearch after being disabled will automatically backfill all indexes, so no data is lost if redisearch is accidentally disabled/deleted on a node.

Note that _by default_ with this update, redisearch is disabled, thus existing nodes that update with this code will have their redisearch disable unless they set the proper helm value when updating.

##### Checklist

### DIFF
--- a/helm/dragonchain-k8s/values.yaml
+++ b/helm/dragonchain-k8s/values.yaml
@@ -8,11 +8,11 @@ global:
     DRAGONCHAIN_ENDPOINT: "https://my-chain.api.company.org:443" # publicly accessible endpoint for this chain. MUST be able to be hit from the internet
     # If using a simple single node cluster like minikube, and the service values below are left as NodePort with port 30000, this would simply be 'http://public.ip.of.minikube.node:30000'
     # Remember to switch the protocol http:// or https:// depending on if you have enabled TLS support or not. (This can be any arbitrary string if no Dragon Net)
-    LEVEL: "1" # Integer 1-5 as a string. Must match with Dragon Net registration if participating in Dragon Net
-    TLS_SUPPORT: "" # Put 'true' here if you have created a TLS certificate and put into a tls kubernetes secret, otherwise leave empty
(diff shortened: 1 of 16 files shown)


helm/dragonchain-k8s/values.yaml
```diff
@@ -8,10 +8,11 @@ global:
     DRAGONCHAIN_ENDPOINT: "https://my-chain.api.company.org:443" # publicly accessible endpoint for this chain. MUST be able to be hit from the internet
     # If using a simple single node cluster like minikube, and the service values below are left as NodePort with port 30000, this would simply be 'http://public.ip.of.minikube.node:30000'
     # Remember to switch the protocol http:// or https:// depending on if you have enabled TLS support or not. (This can be any arbitrary string if no Dragon Net)
-    LEVEL: "1" # Integer 1-5 as a string. Must match with Dragon Net registration if participating in Dragon Net
-    TLS_SUPPORT: "" # Put 'true' here if you have created a TLS certificate and put into a tls kubernetes secret, otherwise leave empty
+    LEVEL: "2" # Integer 1-5 as a string. Must match with Dragon Net registration if participating in Dragon Net
```
rdaquilante: Do we really want to change the default here?

cheeseandcereal: Yeah, I figure most people are installing l2+ anyways, and if they are installing l1, there's extra kubernetes setup they should be aware of. explicitly requiring a user to set '1' just brings more attention to that

scripts/check_packages.py
```diff
@@ -0,0 +1,43 @@
...
+            if line:
+                package = line[: line.find("==")]
+                version = line[line.find("==") + 2 :]
+                if "[" in package:
+                    package = package[: package.find("[")]
+                packages.append((package, version))
+    return packages
+
+
+if __name__ == "__main__":
```
regan-karlewicz: This is awesome.

PR #281, opened 2019-12-20
Add recovery queue and brpoplpush to job processor

rdaquilante:
## Description

Completes an old JIRA item to add recovery flow to job processor in case of unexpected crashes.

##### Checklist

- [x] tests are included

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@
   - Implemented deadlines for L5 blocks based on block times and confirmations for public blockchains
   - Remove any concept of api keys starting with `WEB_` from being special
   - Deprecate support for Ethereum Classic Testnet (Morden) with ethereum interchain functionality
+  - Add retry logic to catch failed job processor jobs
 
 ## 4.3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
   - Fix tests for python 3.8.1
 - **Bugs:**
   - Adds Content-Type header to all verification notifications
+  - Add retry logic to catch failed job processor jobs
 
 ## 4.3.2
 
@@ -48,7 +49,6 @@
   - Implemented deadlines for L5 blocks based on block times and confirmations for public blockchains
   - Remove any concept of api keys starting with `WEB_` from being special
   - Deprecate support for Ethereum Classic Testnet (Morden) with ethereum interchain functionality
-  - Add retry logic to catch failed job processor jobs
 
 ## 4.3.0
 
--- a/dragonchain/job_processor/job_processor.py
+++ b/dragonchain/job_processor/job_processor.py
@@ -120,7 +120,6 @@ def get_next_task() -> Optional[dict]:
     _log.info("Awaiting contract task...")
     pop_result = redis.brpoplpush_sync(CONTRACT_TASK_KEY, PENDING_TASK_KEY, 0, decode=False)
     if pop_result is None:
-        redis.lpop_sync(PENDING_TASK_KEY)
         return None
     _, event = pop_result
     _log.debug(f"received task: {event}")
@@ -129,6 +128,7 @@ def get_next_task() -> Optional[dict]:
         _validate_sc_build_task(event)
     except Exception:
         _log.exception("Error processing task, skipping")
+        redis.lpop_sync(PENDING_TASK_KEY)
         return None
     _log.info(f"New task request received: {event}")
     return event
--- a/dragonchain/job_processor/job_processor.py
+++ b/dragonchain/job_processor/job_processor.py
@@ -40,6 +40,9 @@ STORAGE_LOCATION = os.environ["STORAGE_LOCATION"]
 SECRET_LOCATION = os.environ["SECRET_LOCATION"]
 DRAGONCHAIN_IMAGE = os.environ["DRAGONCHAIN_IMAGE"]
 
+CONTRACT_TASK_KEY = "mq:contract-task"
+PENDING_TASK_KEY = "mq:contract-pending"
+
 _log = logger.get_logger()
 _kube: kubernetes.client.BatchV1Api = cast(kubernetes.client.BatchV1Api, None)  # This will always be defined before starting by being set in start()
 _validate_sc_build_task = fastjsonschema.compile(schema.smart_contract_build_task_schema)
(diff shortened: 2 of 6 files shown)


dragonchain/job_processor/job_processor.py
```diff
@@ -55,6 +58,14 @@ def start() -> None:
     _kube = kubernetes.client.BatchV1Api()
 
     _log.debug("Job processor ready!")
+
+    if redis.llen_sync(PENDING_TASK_KEY):
+        _log.warning("WARNING! Pending job processor queue was not empty. Last job probably crashed. Re-queueing these dropped items.")
+        to_recover = redis.lrange_sync(PENDING_TASK_KEY, 0, -1, decode=False)
+        p = redis.pipeline_sync()
+        p.rpush(CONTRACT_TASK_KEY, *to_recover)
+        p.delete(PENDING_TASK_KEY)
```
cheeseandcereal: this is a race condition if something is added to `PENDING_TASK_KEY` after you check its length, but before you delete it here

rdaquilante: this is copied from queue.py

dragonchain/lib/database/redis.py
```diff
@@ -299,6 +299,23 @@ def brpop_sync(keys: str, timeout: int = 0, decode: bool = True) -> Optional[tup
...
+    """Perform a blocking pop against redis list(s)
+    Args:
+        pop_key: Can be a single key (bytes, string, int, etc), or an array of keys to wait on popping from
+        push_key: key to push currently processing items to
+        timeout: Number of seconds to wait before 'timing out' and returning None. If 0, it will block indefinitely (default)
+    Returns:
+        None when no element could be popped and the timeout expired. This is only possible when timeout is not 0
+        A tuple with the first element being the key where the element was popped, and the second element being the value of the popped element.
+    """
+    _set_redis_client_if_necessary()
```
cheeseandcereal: needs unit test coverage

dragonchain/job_processor/job_processor.py
```diff
@@ -55,6 +58,14 @@ def start() -> None:
     _kube = kubernetes.client.BatchV1Api()
 
     _log.debug("Job processor ready!")
+
+    if redis.llen_sync(PENDING_TASK_KEY):
```
cheeseandcereal: should unit test this branch

dragonchain/job_processor/job_processor.py
```diff
@@ -117,6 +128,7 @@ def get_next_task() -> Optional[dict]:
         _log.exception("Error processing task, skipping")
         return None
     _log.info(f"New task request received: {event}")
+    redis.lpop_sync(PENDING_TASK_KEY)
```
cheeseandcereal: this should probably be somewhere in `start_task`

dragonchain/job_processor/job_processor.py
cheeseandcereal: meh, race condition should actually never happen as long as there is never more than 1 job processor running (which there shouldn't be), so I'll ignore this

dragonchain/job_processor/job_processor.py
```diff
@@ -105,7 +117,7 @@ def get_next_task() -> Optional[dict]:
         The next task. Blocks until a job is found.
     """
     _log.info("Awaiting contract task...")
-    pop_result = redis.brpop_sync("mq:contract-task", 0, decode=False)
+    pop_result = redis.brpoplpush_sync(CONTRACT_TASK_KEY, PENDING_TASK_KEY, 0, decode=False)
     if pop_result is None:
         return None
```
cheeseandcereal: Anywhere this function returns `None` (intentionally throwing away the task), you need to remove the pending task that you just added (since we're _trying_ to intentionally ignore that task)

dragonchain/job_processor/job_processor.py
```diff
@@ -97,6 +108,7 @@ def start_task() -> None:
         if job and (job.status.succeeded or job.status.failed):
             delete_existing_job(task_definition)
         attempt_job_launch(task_definition)
+        redis.lpop_sync(PENDING_TASK_KEY)
```
cheeseandcereal: should pop this off when it does the explicit `return` too, since in that case we're _trying_ to throw away the task; we don't want it coming back

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
   - Implemented deadlines for L5 blocks based on block times and confirmations for public blockchains
   - Remove any concept of api keys starting with `WEB_` from being special
   - Deprecate support for Ethereum Classic Testnet (Morden) with ethereum interchain functionality
+  - Add retry logic to catch failed job processor jobs
```
cheeseandcereal: Rebase your branch and move this line to the `4.3.2` section now

dragonchain/job_processor/job_processor.py
rdaquilante: if `pop_result` is None, doesn't that mean the redis command failed?

cheeseandcereal: `pop_result` being none means the blocking wait timed out (so yeah, i guess  you could call that a failure)

dragonchain/job_processor/job_processor_utest.py
```diff
@@ -59,20 +59,30 @@ def test_get_job_labels(self):
             },
         )
 
-    @patch("dragonchain.job_processor.job_processor.redis.brpop_sync", return_value=(1, valid_task_definition_string))
-    def test_can_get_next_task(self, mock_brpop):
+    @patch("dragonchain.job_processor.job_processor.kubernetes.client.BatchV1Api")
+    @patch("dragonchain.job_processor.job_processor.kubernetes.config.load_incluster_config")
+    @patch("dragonchain.job_processor.job_processor.redis.llen_sync", return_value=True)
+    @patch("dragonchain.job_processor.job_processor.redis")
+    def task_restores_from_pending_queue(self, mock_kube_config, mock_kube_client, mock_redis_llen, mock_redis):
```
cheeseandcereal: this still isn't running....

dragonchain/job_processor/job_processor.py
rdaquilante: if the blocking wait times out, won't that also mean nothing gets added to the second queue?

cheeseandcereal: I guess that's true for this particular branch. so i guess remove the lpop_sync from this branch, but later in this method, there is another `return None` which does need to be fixed

PR #305, opened 2020-01-29
Update L4 block schema and fwatchdog

cheeseandcereal:
## Description

Updates the fwatchdog dependency as well as adding the dragonchain name to the L4 block

When this is released, L1s will have to be updated to accommodate the new L4 block schema

##### Checklist


dragonchain/lib/dto/l4_block_model.py
```diff
@@ -36,7 +36,23 @@ def new_from_at_rest(block: Dict[str, Any]) -> "L4BlockModel":
...
+            timestamp=block["header"].get("timestamp") or "-1",
+            prev_proof=block["header"]["prev_proof"],
+            scheme=block["proof"]["scheme"],
+            proof=block["proof"]["proof"],
+            nonce=block["proof"].get("nonce"),
+            l1_dc_id=block["header"]["l1_dc_id"],
+            l1_block_id=block["header"]["l1_block_id"],
+            l1_proof=block["header"]["l1_proof"],
+            validations=validations,
+            chain_name=block["header"]["chain_name"],
```
regan-karlewicz: Won't this KeyError if the L3 block doesn't contain chain name?

PR #319, opened 2020-05-29
check current level for -1

DRigby26:
## Description

Fix for a bug that was causing failure when broadcast processor couldn't find block

##### Checklist


dragonchain/broadcast_processor/broadcast_processor.py
```diff
@@ -245,6 +245,8 @@ async def process_blocks_for_broadcast(session: aiohttp.ClientSession) -> None:
     for block_id, score in await broadcast_functions.get_blocks_to_process_for_broadcast_async():
         _log.info(f"[BROADCAST PROCESSOR] Checking block {block_id}")
         current_level = await broadcast_functions.get_current_block_level_async(block_id)
+        if current_level == -1:
```
wolfpack94, line 248:
```suggestion
        if current_level != -1:
```

dragonchain/broadcast_processor/broadcast_processor.py
```diff
@@ -245,6 +245,9 @@ async def process_blocks_for_broadcast(session: aiohttp.ClientSession) -> None:
...
+            _log.warning(f"Matchmaking does not have record of claim for block {block_id}.")
```
wolfpack94, line 249:
```suggestion
            _log.warning(f"Failed to lookup current level for block {block_id}.")
```

PR #322, opened 2020-07-16
Subsequent L5 verification search

wolfpack94:
##### Checklist

- [x] tests are included


dragonchain/lib/database/redisearch.py
```diff
@@ -299,23 +304,64 @@ def generate_indexes_if_necessary() -> None:
...
+    block_paths = storage.list_objects("BLOCK/")
+    pattern = re.compile(r"BLOCK\/([0-9]+)-([Ll])5(.*)$")
+    for block_path in block_paths:
+        if LEVEL == "1" and BROADCAST_ENABLED and re.search(pattern, block_path):
+            if not client.redis.sismember(L5_BLOCK_MIGRATION_KEY, block_path):
+                raw_block = storage.get_json_from_object(block_path)
+                block = l5_block_model.new_from_at_rest(raw_block)
+                _log.info("STORING L5 VERIFICATION:")
+                _log.info(block.export_as_search_index())
+                put_document(Indexes.verification.value, block_path.split("/")[1], block.export_as_search_index())
```
cheeseandcereal, line 360: what are the contents of `block_path.split("/")[1]` ?

Remember that regardless of the index name, the document name here must be _globally_ unique in the redisearch

dragonchain/lib/database/redisearch.py
```diff
@@ -299,23 +304,64 @@ def generate_indexes_if_necessary() -> None:
...
+
+def _generate_l5_verification_indexes() -> None:
+    client = _get_redisearch_index_client(Indexes.verification.value)
+    try:
+        client.create_index(
+            [
+                redisearch.NumericField("block_id", sortable=True),
+                redisearch.NumericField("prev_id", sortable=True),
+                redisearch.NumericField("timestamp", sortable=True),
+                redisearch.TextField("dc_id"),
```
cheeseandcereal, line 344: What's the usecase for this field? Textfields are sharded so that you can search with smaller tokens, but I'm assuming you only need this index because you will only ever search on it with the full dc_id, and not part of one.

This should probably be a tagfield in that case because they are lighter than textfields for both storage and querying.

dragonchain/webserver/lib/verifications.py
```diff
@@ -50,6 +52,42 @@ def query_verifications_v1(block_id: str, params: Optional[Dict[str, Any]] = Non
...
 
 
+def query_subsequent_interchain_broadcasts_v1(block_id: str) -> List[Any]:
+    """Return the subsequent broadcasts to other L5 networks"""
+    _log.info(f"Getting subsequent L5 verifications for {block_id}")
+    results = []
+    l5_verifications = _get_verification_records(block_id, 5)
+    if len(l5_verifications) > 0:
+        timestamp = cast(List[Any], l5_verifications)[0]["header"]["timestamp"]
+        l5_nodes = redisearch._get_redisearch_index_client(redisearch.Indexes.verification.value).redis.smembers(redisearch.L5_NODES)
```
cheeseandcereal, line 62: why are you getting a set of L5 nodes and doing N queries?
Isn't the whole point of the verification index that they're _only_ L5 blocks anyways? It seems like you would only need one query if that were the case.

dragonchain/webserver/lib/verifications.py
```diff
@@ -50,6 +52,42 @@ def query_verifications_v1(block_id: str, params: Optional[Dict[str, Any]] = Non
...
+        query_result = redisearch.search(
+            index=redisearch.Indexes.verification.value,
+            query_str=f"@dc_id:{l5_dc_id} @timestamp:[{int(timestamp) + 1} +inf]",
+            only_id=True,
+            limit=1,
+            sort_by="timestamp",
+            sort_asc=True,
+        )
+    except redis.exceptions.ResponseError as e:
+        error_str = str(e)
```
cheeseandcereal, line 79: I don't think any of this additional error handling is relevant here. The index in your query is hardcoded and you're not passing user input into the query, so it should never have a syntax error. Probably just remove this entire try/except

dragonchain/webserver/routes/verifications.py
```diff
@@ -39,3 +42,8 @@ def get_verification_v1(block_id: str, **kwargs) -> Tuple[str, int, Dict[str, st
 @request_authorizer.Authenticated(api_resource="verifications", api_operation="read", api_name="get_pending_verifications")
 def get_pending_verifications_v1(block_id: str, **kwargs) -> Tuple[str, int, Dict[str, str]]:
     return helpers.flask_http_response(200, verifications.get_pending_verifications_v1(block_id))
+
+
+@request_authorizer.Authenticated(api_resource="verifications", api_operation="read", api_name="get_subsequent_verifications")
```
cheeseandcereal, line 47: `get_subsequent_verifications` will need to be added to the endpoint map in api_key_model, the permission document schema, and the table in the documentation on the permissioning page.

dragonchain/webserver/lib/dragonnet.py
```diff
@@ -87,6 +89,10 @@ def process_receipt_v1(block_dto: Dict[str, Any]) -> None:
                     matchmaking.add_receipt(l1_block_id, level_received_from, dc_id, block_id, proof)
                 except Exception:
                     _log.exception("matchmaking add_receipt failed!")
+                if level_received_from == 5:
+                    client = redisearch._get_redisearch_index_client(redisearch.Indexes.verification.value)
+                    client.redis.sadd(redisearch.L5_NODES, block_model.dc_id)
+                    block_dao.insert_l5_verification(storage_location, block_model)
```
cheeseandcereal, line 95: This should definitely only happen after the broadcast system processes this block, since the received verification in the broadcast system can be rejected, in which case you definitely don't want to save this.

cheeseandcereal:
Should probably add some tests

dragonchain/lib/database/redisearch.py
wolfpack94: the block path in this case is `BLOCK/<block_id>-L5-<l5_dc_id>` so I only want this part `<block_id>-L5-<l5_dc_id>`

dragonchain/lib/database/redisearch.py
wolfpack94: thanks man. Thats a good point. I only needed that one extra field so good call

dragonchain/webserver/lib/dragonnet.py
wolfpack94: great point

dragonchain/webserver/routes/verifications.py
```diff
@@ -39,3 +40,8 @@ def get_verification_v1(block_id: str, **kwargs) -> Tuple[str, int, Dict[str, st
 @request_authorizer.Authenticated(api_resource="verifications", api_operation="read", api_name="get_pending_verifications")
 def get_pending_verifications_v1(block_id: str, **kwargs) -> Tuple[str, int, Dict[str, str]]:
     return helpers.flask_http_response(200, verifications.get_pending_verifications_v1(block_id))
+
+
+@request_authorizer.Authenticated(api_resource="verifications", api_operation="read", api_name="query_interchain_verifications")
```
cheeseandcereal, line 45: still needs to be added to permissions doc schema in schema.py and table in docs/usage/permissioning.md

wolfpack94: thanks. I definitely missed those

